PR #23, opened 2026-06-03
fix(substrate): handle wrapped namespace table-not-found errors

tenequm:
## Summary
- detect wrapped Lance namespace TableNotFound errors by walking typed error sources
- let open_or_create_via_ns create missing tables instead of failing on wrapped namespace errors

## Verification
- cargo check


tenequm:
Superseded by #22 - the same fix was cherry-picked as 19847a1 and cleaned up in d10517f (collapsed the two-fn recursion into a single std::iter::successors walker, added unit test for wrapped/direct/mismatched/non-namespace paths). Already merged to main via #22.

PR #87, opened 2026-07-03
ci: temp dist artifact check (do not merge)

tenequm:
Temporary: exports dist/ as an artifact for local prerelease execution checks. Close without merging.


tenequm:
Prerelease checks done - artifacts verified locally.

Issue #75, opened 2026-06-30
pond_search from_date/to_date returns empty (ZoneMap mis-prunes tz-aware timestamp column)


tenequm:
## Summary

`pond_search`'s `from_date` / `to_date` filters return (almost) nothing when the store is on remote S3, even though the matching data is present. Any date bound collapses the in-scope set to ~0 rows - including a bound far in the past that should match everything. Date-scoped recall ("last N days") therefore silently returns an empty result, masking the full corpus.

This is a correctness bug, not a performance one. It can be fixed and shipped on its own.

> **Root cause corrected (2026-06-30).** The original hypothesis below ("tz-naive literal vs tz-aware column, fix the literal") was **wrong**. Empirical testing showed the literal form is irrelevant: *no* literal form works while the `messages_timestamp_zonemap` index is present, and *every* literal form works once it is gone. The real cause is the ZoneMap scalar index mis-pruning a timezone-aware timestamp column. Corrected analysis and fix are in the sections below.

## Evidence

Store: `s3+https://nbg1.your-objectstorage.com/pondarium/pond`. Project scope: `/Users/tenequm/Projects/pond`.

| Query | in-scope | hits |
|---|---|---|
| `pond_search` fts/vector, no date filter | 33,539 | real |
| `pond_sql_query` `WHERE timestamp >= '2026-06-16'` (non-subagent) | 17,221 | (DataFusion, correct) |
| `pond_search` fts/vector, `from_date=2026-06-16` | ~0 | none |
| `pond_search` fts, `from_date=2020-01-01` (matches everything) | 0 | none |

Reproduced locally against the fixture corpus (which builds the same zonemap), via `Dataset::count_rows(Some(filter))` on the `messages` table (1574 rows total):

| filter | with zonemap | without zonemap |
|---|---|---|
| `timestamp IS NOT NULL` | 1574 | 1574 |
| `timestamp >= timestamp '2000-01-01 00:00:00'` (matches all) | **0** | 1574 |
| `timestamp <  timestamp '2099-01-01 00:00:00'` (matches all) | **0** | 1574 |
| `timestamp >= timestamp '2000-01-01 00:00:00+00:00'` (offset) | **0** | 1574 |
| `timestamp >= cast('2000-01-01 00:00:00' as timestamp(0/3/9))` | **0** | 1574 |
| `timestamp >= arrow_cast('2000-01-01T00:00:00Z', 'Timestamp(Microsecond, Some("UTC"))')` | **0** | 1574 |

[...]

PR #100, opened 2026-07-07
feat(parts)!: materialize tool columns + in-place schema migration; per-query SQL timeout

### LINKED ISSUE #89: Remote-store SQL analytics over parts.variant_data time out: compress JSON columns + materialize tool columns
## Symptom

On a remote (S3) store, corpus-wide tool-call analytics in `pond_sql_query` hit the 30s timeout guard (`QUERY_TIMEOUT`, `src/sql.rs`). This includes the schema doc's own flagship worked example ("tool usage and failure rates over the last week", `src/transport.rs` schema resource), which fails even scoped to 7 days. Even a plain `GROUP BY json_get_string(variant_data, 'name')` over all tool_call parts times out.

Found during an MCP stress test (2026-07-03) against the real corpus: 12,330 sessions / 2.2M messages / ~1.5M parts on a Hetzner object store.

## Measurements

Same queries, same corpus, remote (MCP over S3) vs local reference store (`pond sql --storage-path local`):

| Query | S3 | Local |
|---|---|---|
| `GROUP BY json_get_string(variant_data,'name')` over all 490K tool_calls | timeout >30s | 1.9s |
| Flagship failure-rate self-join, 7-day scope | timeout >30s | 4.0s |
| `COUNT(*) WHERE type='tool_call'` (no JSON column touched) | 0.7s | instant |

So the timeout is pure network I/O, not compute: any JSON extraction reads the `variant_data` column bytes, and the query cannot finish pulling them from the object store inside 30s.

Column sizes (logical JSON text, measured):

| type | rows | avg chars | total |
|---|---|---|---|
| tool_result | 488K | 4,218 | 1.96 GB |
| text | 328K | 1,132 | 354 MB |
| tool_call | 488K | 590 | 275 MB |
| reasoning | 201K | 632 | 121 MB |

Logical total ~2.7 GB vs `parts.lance` data dir 2.5 GB on disk: **ratio ~1.0, i.e. zero effective compression**. Because tool_call rows are interleaved with fat tool_result rows in the same pages, late materialization still pulls essentially the whole column even when only tool_call bodies are needed.

## Root cause

`variant_data` is JSONB stored as `LargeBinary`, and Lance 7.0.0's default encoding strategy leaves it raw:

- The auto-FSST path explicitly excludes `Binary`/`LargeBinary` (`lance-encoding-7.0.0/src/compression.rs:492`).
- General compression (zstd/lz4) applies only when explicitly configured via `lance-encoding:compression` field metadata; pond sets none (`json_field()` in `src/sessions.rs`).

Meanwhile `messages.search_text` (Utf8) does get auto-FSST - only the JSON columns ship uncompressed.

Secondary factors:

- `parts` has no timestamp column, so the date filter (on `messages.timestamp`) cannot prune the JSON extraction before the join.
- No index helps these shapes: Lance 7.0.0's JsonIndex only accelerates pushed-down equality predicates (`json_get_string(col,'name') = '...'`), never GROUP BY extraction or join-key extraction. The FM-Index plan (#47) targets substring hunts, a different problem.

### Lever 1: zstd on the JSON columns (substrate fix)

Add `lance-encoding:compression = "zstd"` field metadata to `variant_data` and `options` in `part_schema()` (and the message/session schemas). Lance wraps the encoder in `GeneralMiniBlockCompressor` per chunk; decode is chunk-local, so indexed point reads (pond_get hydration) do not pay a full-column penalty.

Impact: JSON tool bodies compress ~4-8x, so the 2.5 GB scan becomes ~350-600 MB. Current numbers imply ~80 MB/s effective throughput, putting the compressed corpus-wide `json_get_*` scan at roughly 5-8s - inside the 30s cap. Also shrinks hydration, `pond copy`, and sync reads over remote by the same factor, plus storage cost.

Migration without any local store: `Dataset::replace_field_metadata` exists in Lance 7.0.0 (`lance-7.0.0/src/dataset.rs:3270`) - flip the metadata on the live remote dataset, then compaction (`pond optimize`) rewrites fragments re-encoded. Needs a scratch-store verification that compaction picks up the new metadata; fallback is a fresh ingest (adapter sources always live on the client machine, so no local store dependency).

### Lever 2: materialize `tool_name` / `call_id` / `is_failure` as native nullable columns on `parts`

Compression makes the full-column scan pass but it still scales linearly with corpus size; the documented analytics should never touch the fat column at all.

- Three narrow nullable columns (`Utf8`, `Utf8`, `Boolean`), ~30-50 MB total at current scale, filled at ingest. Trivial because the data is already typed at the seam (`PartKind::ToolCall`/`ToolResult` in `src/wire.rs`); absent stays NULL, consistent with the no-synthesis rule.
- Write path: `part_schema()` +3 fields, `parts_chunk()` +3 arrays. Read path unchanged (`variant_data` stays the verbatim source of truth). The SQL parts view builds its projection dynamically from the dataset schema (`src/sql.rs::register`), so the new columns appear automatically. Copy/sync are schema-driven, nothing to touch.
- With `tool_name` + `is_failure` present on tool_result rows, the flagship failure-rate query loses its call_id self-join entirely - it becomes a single-table GROUP BY over narrow columns, the same class as the `type` count that ran in 0.7s against S3.
- Optional: bitmap index on `tool_name` via the existing `PARTS_SCALAR_INDICES` machinery (preferable to a JsonIndex, which cannot serve these shapes anyway).
- Schema change is breaking (`feat!:`), fine pre-release; stores rebuild via re-ingest + `pond copy` (append fast-path, 13.8 min measured for the full corpus) or the lever-1 migration path.

### Also

- Update the `schema://pond-sql` resource: rewrite the flagship example to the native-column form once lever 2 lands.
- Keep `QUERY_TIMEOUT` at 30s - after both levers it guards genuine abuse again instead of walling off the documented surface.
- Blob v2 payload separation (spec deferred item 11, blocked on the upstream compact bug) remains the eventual structural endgame; not needed once these land.

## Expected results (remote store)

| Query shape | Today | After |
|---|---|---|
| Tool usage / failure rates (documented examples) | timeout >30s | ~1-3s, corpus-growth-proof |
| Arbitrary `json_get_*` over all tool bodies | timeout >30s | ~5-10s (compressed scan) |
| pond_get hydration, copy, sync remote reads | full-size | ~4-8x fewer bytes |

tenequm:
# feat(parts)!: materialize tool columns + in-place schema migration; per-query SQL timeout

Closes #89.

## What

Remote-store tool analytics over `parts.variant_data` (multi-GB uncompressed JSONB) timed out hard: any JSON-touching query pulled the whole column over S3. This PR makes the common analytics shapes never touch the fat column:

- **3 derived storage columns on `parts`**: `tool_name` (Utf8), `call_id` (Utf8), `is_failure` (Boolean) - all nullable, extracted from the already-typed `PartKind` at write time (no-synthesis: absent stays NULL). `variant_data` remains the verbatim source of truth.
- **BTree scalar index on `tool_name`** (BTree, not Bitmap: prefix LIKE errors on bitmap indexes upstream).
- **Flawless in-place migration**: existing stores upgrade on first open - `classify_schema` detects the missing nullable columns and a one-time `add_columns` backfill derives them from stored `variant_data` (~seconds locally, no re-ingest, spec.md#session-durable-copy holds). Pre-#89 `.pond` archives restore via read-boundary derivation; the archive file is never modified. Stores written by a NEWER pond open read-only-compatibly with a warning (unknown extra columns).
- **Spec 5.6** (`session-additive-schema-backfill`): the derived-column class and the migration contract are now spec rules.
- **Per-query SQL timeout**: `timeout_seconds` on `pond_sql_query`, `--timeout` on `pond sql` (default 30s, clamp 1..600). The timeout error names the knob and steers to the native columns.
- SQL schema doc, tool description, and error hints rewritten around the native columns (flagship example no longer uses JSON getters).

## Explored and rejected: zstd on the JSON columns (Lever 1)

Measured 0.0% compression on the real corpus in 3 configs: lance miniblock zstd compresses 4 KiB chunks independently, and real JSON bodies carry cross-ROW redundancy, not within-body (raw same-bytes full-window zstd: 4.8-7.4x). `minichunk=32768` is silently rejected by the v2.1 `>= 32*1024` guard. Values >32 KiB are already auto-compressed by lance v8 per-value. WHY comment lives on `json_field()`; do not re-explore blind. (Miniblock chunk size becomes tunable in lance v9 - re-evaluate then.)

## Numbers (full real corpus)

Remote = `pond-bench-89`, 10,840 sessions / 1.81M messages / ~608K tool_call parts (28% MORE tool calls than the pre-#89 baseline store). Baseline = pond 0.12.2 on the pre-#89 copy.

| Query (remote S3) | Baseline | This PR |
|---|---|---|
| tool GROUP BY (JSON getter) | **TIMEOUT >30s** | 29.3s / TIMEOUT (same store - old form still dies) |
| tool GROUP BY (native `tool_name`) | - | **8.5-9.7s** |
| flagship failure-rate self-join 7d (JSON getters) | **TIMEOUT >30s** | **TIMEOUT** (reproduced) |
| flagship (native columns) | - | **23-24s** (one cold 32s outlier) |
| indexed point filter `tool_name='Bash'` | - | 4.7-8.0s |
| narrow COUNT control | 1.8s | 3.2-4.9s (larger corpus) |

Local: tool GROUP BY 1,693ms -> **48ms** (~35x); flagship 3,138ms -> **220ms** (~14x).

Residual (isolated, pre-existing, NOT this PR's scope): the 7-day scoping alone costs 11.8s - `messages.timestamp` has no index; every scalar index type funnels literals through the shared `safe_coerce_scalar` tz-dropping coercion (#75), so a timestamp index is blocked upstream. Unscoped flagship: 9.9s; sessions-scoped: 18.5s. The new `timeout_seconds` gives callers headroom for such shapes.

## No degradation (regression sweep)

- ingest_bench: profile identical (0.14s wall, same stage split) despite the 3 extra arrays per parts chunk.
- read_bench: within 1-2ms noise at every tail size.
- serve_mem on the new-schema corpus: FTS p50 7ms / vector 24ms / get 34ms / sql 10ms; idle 493.6 MiB and peak 1683 MiB PF - both targets PASS.
- Real-corpus S3 copy with the new schema: 40:28 total (upload ~14 min, index rebuild 26:21, verify 32s), `SYNCED - 2,933,838 rows, 0 duplicates`.
- Scalar-fold A/B on S3 (bonus, lance-8 delta segments): folding scalar indexes every sync costs ~6-10s of S3 round trips per sync (67s outlier) vs 0 deferred - the 50k deferral gate stays correct under v8; its comment rationale (whole-file rewrite) is stale and can be refreshed in a follow-up.

## Validation

- `cargo fmt --check` clean, `cargo clippy --all-targets -- -D warnings` clean, `cargo test` **272 passed / 0 failed**.
- New integration suite `schema_migration`: in-place backfill (values match ingest-time derivation, non-tool parts stay NULL, second open commits nothing, `tool_name` index builds on the migrated store via standard intents) + archive restore derivation (archive version byte-untouched).
- Migration exercised on a real store shape: seed -> raw-lance `drop_columns` -> reopen -> backfill notice -> correct values.

## Breaking

`feat!`: stores written by this version are unreadable by pond <= 0.12.2 (strict schema equality check in old binaries). Old stores upgrade automatically and silently on first open by a new binary; the only visible artifact is a one-time stderr notice. Production upgrade path: ship binary, first open migrates in place (backfill measured in seconds locally; remote pays one `add_columns` commit).

Do not merge until reviewed.

### DIFF
--- a/benches/serve_mem_bench.rs
+++ b/benches/serve_mem_bench.rs
@@ -662,7 +662,7 @@ async fn run_sql_phase(
             messages: Some(store.dataset(Table::Messages).await?),
             parts: Some(store.dataset(Table::Parts).await?),
         };
-        match sql::run(&tables, q, Mode::Inline, sql::DEFAULT_INLINE_ROWS).await {
+        match sql::run(&tables, q, Mode::Inline, sql::DEFAULT_INLINE_ROWS, None).await {
             Ok(_) => {}
             Err(error) => anyhow::bail!("{name}: sql {q:?} failed: {error:?}"),
         }
--- a/benches/sync_oracle_bench.rs
+++ b/benches/sync_oracle_bench.rs
@@ -128,12 +128,18 @@ async fn fetch_tables(store: &Store) -> Result<Tables> {
 /// uses. Returns the row count from the result; the inline rendering cost is
 /// kept tiny by passing `inline_rows = 0`.
 async fn run_sql(tables: &Tables, sql: &str) -> Result<usize> {
-    let outcome = pond::sql::run(tables, sql, Mode::Export(pond::sql::Format::Ndjson), 0)
-        .await
-        .map_err(|err| match err {
-            pond::sql::SqlError::Query(msg) => anyhow::anyhow!("query: {msg}"),
-            pond::sql::SqlError::Infra(err) => err,
-        })?;
+    let outcome = pond::sql::run(
+        tables,
+        sql,
+        Mode::Export(pond::sql::Format::Ndjson),
+        0,
+        None,
+    )
+    .await
+    .map_err(|err| match err {
+        pond::sql::SqlError::Query(msg) => anyhow::anyhow!("query: {msg}"),
+        pond::sql::SqlError::Infra(err) => err,
+    })?;
     let count = match outcome {
         Outcome::Inline(_) => 0,
         Outcome::Export { rows, .. } => rows,
--- a/benches/tokenizer_quality_bench.rs
+++ b/benches/tokenizer_quality_bench.rs
@@ -309,7 +309,7 @@ async fn resolve_anchors(store: &Store, queries: &mut [Query]) -> Result<()> {
             let escaped = anchor.replace('\'', "''");
             let sql =
                 format!("SELECT message_id FROM messages WHERE search_text LIKE '%{escaped}%'");
-            match sql::run(&tables, &sql, Mode::Export(sql::Format::Ndjson), 0).await {
+            match sql::run(&tables, &sql, Mode::Export(sql::Format::Ndjson), 0, None).await {
                 Ok(sql::Outcome::Export { bytes, .. }) => {
                     for line in String::from_utf8_lossy(&bytes).lines() {
                         if line.is_empty() {
--- a/docs/site/src/pages/guides/several-machines.mdx
+++ b/docs/site/src/pages/guides/several-machines.mdx
@@ -5,3 +5,7 @@ Point each machine's `config.toml` (or `POND_*` env) at the same URL and credent
 Concurrent writers are safe: Lance serializes commits through the object store's conditional writes (optimistic concurrency), so two syncs at once cannot corrupt the dataset.
 
 Within one machine, sync is single-flight: a second `pond sync` against the same store waits for the first (naming the holder), and the scheduled run passes `--no-wait` so ticks skip cleanly instead of queueing. `pond status --hosts` shows which machines feed the store - sessions and latest activity per ingest host.
+
+## Mixed versions
+
+A release that changes the store schema (0.13.0 added materialized tool columns) migrates the store in place the first time the new binary opens it - a one-time backfill with a stderr notice - after which older binaries fail at open with a schema-mismatch error. When several machines share one bucket, upgrade pond on all of them together; a machine left on an older version stops syncing until it upgrades.
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -362,7 +362,7 @@ This section is how the canonical model of Section 4 persists on the substrate o
 
 ### 5.1 Three datasets
 
-The sessions consumer registers three Lance tables: `sessions`, `messages`, and `parts`. Each is a direct serialization of its canonical type - no projections, no promotions - except that `messages` additionally carries a message's derived embedding (5.5).
+The sessions consumer registers three Lance tables: `sessions`, `messages`, and `parts`. Each is a direct serialization of its canonical type plus a small, named set of derived storage columns with no canonical counterpart: `messages` carries the message's derived embedding (5.5), and `parts` carries the materialized tool-identity columns (5.6). Nothing else is projected or promoted.
 
 `sessions` - one row per Session:
 
@@ -398,6 +398,9 @@ The sessions consumer registers three Lance tables: `sessions`, `messages`, and 
 | `ordinal` | position within the message's content |
 | `type` | the Part discriminator |
 | `provenance` | conversational vs injected (`model-part-provenance`, 4.8); search reads it to exclude injected scaffolding |
+| `tool_name` | derived tool-identity column (5.6); scalar-indexed - tool analytics run on the narrow columns instead of scanning `variant_data` |
+| `call_id` | derived tool-call correlation id (5.6) |
+| `is_failure` | derived ToolResult flag (5.6); non-null only on `tool_result` rows |
 | `variant_data` | JSON (Lance `pa.json_()`, stored as JSONB); the variant-specific fields |
 | `data` | Lance blob; FilePart payload only |
 | `options` | JSON (Lance `pa.json_()`, stored as JSONB) |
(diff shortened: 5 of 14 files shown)


tenequm (commented):
## Review: multi-agent pass (correctness, convention, efficiency/safety)

(GitHub blocks approving one's own PR; verdict is APPROVE with the fixes below applied on-branch.)

**Automated checks:** `cargo fmt --check` clean, `cargo clippy --all-targets -- -D warnings` clean, `cargo test` 272 passed / 0 failed.

### Significant

1. `src/sessions.rs:4964` - [robustness] The backfill mapper propagates a `PartKind` decode error with `?`, failing the whole `add_columns` migration - and since migration re-runs on every open until it succeeds, one undecodable `variant_data` row makes the store permanently un-openable. The read path (`part_from_batch`, same decode) breaks only that one `pond_get`. Pond-written `variant_data` always round-trips (all required fields are unconditionally serialized), so this needs genuine corruption or a future required field - low probability, high blast radius. Fix: per-row decode failure degrades to NULL cells + a warn naming the row (spec 5.6 supports it: "a cell the stored record cannot justify stays NULL").

### Minor

1. `tests/integration/schema_migration.rs` - [test coverage] The suite never writes new tool parts into the migrated store, whose on-disk column order differs from `part_schema()` (`add_columns` appends at the end). Verified lance 8.0.0 appends align by name (`SchemaCompareOptions { ignore_field_order: true, .. }`, vendored `insert.rs:310-318`), so no bug expected - but a post-migration ingest assertion pins that guarantee against future lance bumps.
2. `src/sessions.rs:4931-4946` - [convention] The `column_backfill`/`parts_backfill` bail messages name only the symptom; every sibling error added by this PR names the fix.
3. `src/sessions.rs:5816-5822` - [convention] `tool_identity` doc comment runs 3 sentences; project cap is two.
4. `src/substrate.rs` (`ensure_current_schema`) - [cosmetic] The `UnknownExtra` arm returns early and skips the embedding-dim sanity warning the `Match` arm runs.

### Notes (by design, not defects)

- First open after the upgrade pays the full remote backfill (streams `type` + `variant_data`; minutes on an object store) - whichever process opens first blocks for it; stderr notice exists.
- Newer-pond stores now open read-compatibly (warn) instead of failing at open; the failure moves to write time. Intentional forward-compat trade.

Everything else checked clean: steady-state open adds zero network cost (pure in-memory schema diff); backfill is column-projected, exactly one commit; ingest re-parses no JSON; the timeout wraps the full `collect()`, is capped at 600s, and cannot bypass the read-only gate; OCC races and mid-migration interrupts are safe; test placement, naming, spec-section style, and helper reuse all match existing precedent.

All findings are being fixed on this branch.

Issue #71, opened 2026-06-21
Embedding writeback merges derived vectors into the canonical table (decide: C/D/E)


tenequm:
## Summary

`pond sync`'s embed stage writes vectors back into `messages` via `merge_update` on the schema primary key `(session_id, id)`. On the real 2.1M-row local store, updating **8 vectors takes 6.36s and reads 143 MiB** - the merge *join* scanning the full `session_id`+`id` key columns to locate the rows. The write itself is tiny (1 data file + 1 deletion vector); the join is the cost. Lance 7.0.0 only index-accelerates **single-column** merge keys (`merge_insert.rs::join_key_as_scalar_index` returns `None` when `on.len() != 1`), so pond's composite key gets no index path - a full key scan on every embed. The cost is fixed per sync regardless of how few rows are new, so it dominates every active-coding sync.

## Root cause

The `vector` is lazily-derived, separately-written data stored as a **mutable column in the append-once `messages` table**. That forces an UPDATE, and in a columnar immutable-fragment store an update is the expensive operation. The same root produced three costs we've so far treated separately:

- the 6.36s composite-key merge join (this issue);
- embedding's `merge_update` rewriting message fragments, which churns the resident rowmap (mitigated by the rowmap delta-extend, but the churn remains);
- `vector IS NULL` scanning the 1.2 GB Float16 column (mitigated by the co-set `embedding_model IS NULL` swap).

The cure is to make embedding an **append, not an update**.

## The constraint that makes this non-trivial

pond's vector arm does **scalar-prefilter pushdown** - it prefilters the kNN by `project`/date/`session_id` in the same scan over `messages` (`spec.md#search-prefilter-pushdown`, the load-bearing fix for remote vector-arm latency). Separating the vector out means the filter columns are no longer co-located with the vector, so you either denormalize those columns into the vector store or lose pushdown. That tension - write-path win vs read-path regression - is the actual decision.

### C - separate append-only vector dataset
Vectors in their own Lance dataset keyed by `(session_id, id)`; IVF_SQ index moves there; embedding becomes an `append_stream` (bandwidth-bound, no join, no fragment rewrite). Search runs kNN there, hydrates metadata via the existing row-key map.
- **Pro:** kills the merge join AND the rowmap churn AND removes the 1.2 GB column from `messages`. Best fit for the imminent S3 backend (append is bandwidth-bound; merge is commit-latency-bound on S3 - 5.47x in the copy benchmark).
[...]

PR #104, opened 2026-07-09
Feat/no embed sync and watch

abhishekt-hydra:


abhishekt-hydra:
wrong branch.

tenequm:
@abhishekt-hydra thanks - there's good material here, but it needs to land as separate PRs. How I'd split it:

1. **Multi-root adapters** - welcome as a standalone PR as-is; the `config_roots` seam work looks solid.
2. **`pond share`** - open an issue first to settle the `[share]` config shape and creds scoping, then a PR.
3. **`pond watch`** - issue first: a resident daemon + three watcher dependency trees needs a case for what the 5-min `pond schedule` tick doesn't cover.
4. **No-embed-by-default** - please open an issue for this one; it needs deliberate consideration, because it replays a design pond already shipped and spent three PRs digging out of (#71, #78, #82). When embedding was a deferred backlog, every remote sync paid for that backlog whether or not anything needed embedding. Measured on the real S3 corpus (~2.1M messages):

   | remote `pond sync` wall time | |
   |---|--:|
   | deferred-embed design (backlog-based) | **80-520 s** |
   | inline embed at ingest (current) | **~44 s** |

   Where the deferred design's time went, per sync (#82): **6-81 s** in the embed stage - mostly a full S3 scan to discover the backlog, usually finding nothing - plus **6-27 s** scanning the `embedding_model` column just to report coverage. And actually *filling* a backlog is worse than it looks: writing just **8 vectors took 6.36 s and read 143 MiB** (#71), because the composite-key merge has no index path and must scan the full key columns - a cost that grows with corpus size, not batch size.

   Against that, what skipping embed at ingest saves: sub-second compute for a typical sync delta (candle on Metal or CPU - no GPU; a caught-up sync never loads the model). The current ~44 s sync is only possible *because* inline embed guarantees the backlog is empty, letting sync skip those scans entirely - a no-embed default makes the backlog the permanent state and brings them back. If the driving case is `pond watch` wanting minimal per-tick latency, let's quantify in the issue what inline embed actually adds per debounced tick.

Also: pond is spec-driven - check `docs/spec.md` before changing behavior (this default is spec rule `session-embed-from-canonical`, sections 5.5/8.6); single-feature PRs citing the relevant spec rules move much faster.

PR #124, opened 2026-07-24
fix(substrate): prevent byte-capped compaction loops

### LINKED ISSUE #123: Compaction can rewrite byte-capped fragments on every sync
## Summary

Pond can repeatedly rewrite the same live fragments when the row target is larger than the number of rows that can fit under the configured byte cap. Each rewrite completes successfully, but its outputs remain below the candidacy threshold, so a later small append makes the same output set eligible again.

I observed this on Pond v0.13.2. The relevant code is unchanged on current `main` after v0.14.2.

## Evidence from the affected store

The latest live `parts` layout contained:

- 105,087 rows
- 1,027,449,798 live data bytes
- 9,777 average bytes per row

Pond's byte-derived half-cap target was 13,727 rows, but the fixed lower clamp raised it to 50,000 rows. At the observed row width, a 256 MiB output can hold about 27,455 rows. A compacted output therefore cannot reach the 50,000-row candidacy threshold.

The retained manifests show 31 near-full rewrites of essentially the same live layout. Those rewrites created 30,935,759,950 replacement bytes. The `parts` table reached 32,027,432,975 physical bytes while its current live files needed about 1.03 GB.

The fixed floor and byte cap are set in [`substrate.rs`](https://github.com/tenequm/pond/blob/46cfc95e9c146d77734c45eba5d89f633ee437e5/packages/pond/src/substrate.rs#L844-L846), and the target calculation clamps to that floor [here](https://github.com/tenequm/pond/blob/46cfc95e9c146d77734c45eba5d89f633ee437e5/packages/pond/src/substrate.rs#L1006-L1032).

Lance uses `target_rows_per_fragment` to decide candidacy, while `max_bytes_per_file` only controls how selected fragments are rewritten. Its v8.0.0 source documents that split in [`CompactionOptions`](https://github.com/lance-format/lance/blob/v8.0.0/rust/lance/src/dataset/optimize.rs#L157-L177), applies the row threshold in the planner [here](https://github.com/lance-format/lance/blob/v8.0.0/rust/lance/src/dataset/optimize.rs#L664-L730), and passes both limits to the writer [here](https://github.com/lance-format/lance/blob/v8.0.0/rust/lance/src/dataset/optimize.rs#L1614-L1626).

## Expected behavior

Automatic compaction should converge. After a compaction task completes, a tiny append should not cause all of that task's outputs to be selected and rewritten again.

A one-pass fragment-count reduction is not enough. For example, five byte-capped inputs can become four byte-capped outputs while all four outputs remain below the row threshold. The next small append can restart the same rewrite.

## Proposed fix

Keep this fix at the shared compaction seam:

1. Remove the fixed 50,000-row lower bound and retain only the one-row safety floor.
2. Before executing each exact Lance-planned non-deletion task, require its observed row and byte shape to make the row target attainable under the byte cap. Missing size metadata should fail safe by skipping the optional rewrite.
3. Keep deletion-materialization tasks eligible because their purpose is to remove tombstoned rows.
4. Add an incident-shaped regression and a second-cycle regression that proves the prior outputs no longer all qualify after a tiny append.

Cleanup retention made the amplification visible on disk, but it did not create the replacement bytes. Cleanup cadence and low-disk recovery should be handled separately from this convergence fix.

alexnayko:
Closes #123

## Root cause

Lance selects compaction candidates by row count, while re-encoding can split output at a soft byte limit. Wide-row output can remain below the row target and become a candidate again on the next sync.

The incident table had 1,027,449,798 data-file bytes across 105,087 physical rows, or about 9,777 bytes per row. Pond derived a half-budget target of 13,727 rows, then raised it to the fixed 50,000-row floor. That target could not fit inside the configured 256 MiB output budget.

A one-pass fragment-count reduction does not prove convergence. A five-fragment task can become four byte-capped fragments and then rewrite those four on every later sync.

## Fix

- Derive the row target from observed physical row width without a fixed row floor.
- Require every optional compaction task to strictly reduce fragment count, based on `ceil(total task bytes / max_bytes_per_file)`.
- When a task exceeds one output file's byte budget, require every input fragment's observed physical-row width to let an output reach the row target within half that budget.
- Fail closed when file sizes or the configured byte budget are unavailable.
- Keep deletion materialization eligible. Preserve the 64-fragment backstop and the existing absorb veto after the strict-shrink check.
- Log the skip reason as `missing_sizes`, `invalid_byte_budget`, `cannot_shrink`, `row_target_unattainable`, or `absorb_veto`.

This adds no CLI command, dependency, migration, cleanup behavior, or scheduler behavior.

## Regression coverage

Unit coverage includes the measured incident target, the 5 -> 4 -> 4 rewrite loop, strict-shrink behavior, missing sizes, deletion precedence, physical-row accounting after deletions, the 64-fragment backstop, productive peer merges, and the absorb veto.

The kept cases include both real mixed-width tasks from review and an above-budget off-boundary task. That last fixture derives its target from the whole table while filtering one planned task, so it exercises the conditional width gate without relying on exact boundary equality.

## Validation

- On code head `b7a1727`, Rust 1.95 passed `cargo fmt`, `cargo clippy --locked --offline -- -D warnings`, and the full test suite with 363 passed and 1 ignored.
- The test-only follow-up `e74de1f` passes Rust 1.95 `rustfmt --check`, `git diff --check`, and an independent exact-integer check of the new fixture. Final-head CI is [run 30198684977](https://github.com/tenequm/pond/actions/runs/30198684977).
- A normal optimize canary on the affected store reduced active `parts` fragments from 14 to 6. An immediate second pass made zero file changes. Row counts, corpus hashes, protected transcript hashes, and the FTS result remained identical.

The canary validated end-to-end convergence on the affected store. Its productive task fit below one output budget, so the above-budget width branch is covered by the focused unit fixture.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -185,6 +185,10 @@ An unenforced primary key on the primary-key columns, so merge-insert defaults t
 
 All of a consumer's datasets share one Lance cache and one object-store client. Why: one pool, rather than one per table, avoids multiplying connections and credential refreshes on object-store backends.
 
+#### `lance-compaction-filter`
+
+Automatic compaction runs only Lance-planned tasks that pass pond's filter. Tasks above the deletion-materialization threshold always run - reclaiming tombstones is useful even when layout does not improve. Any other task must strictly reduce the fragment count, and when its total bytes exceed one output file's byte budget, every input fragment's observed row width must let an output reach the row target within half that budget (headroom because re-encoding can widen rows relative to the input estimate). Missing file sizes fail closed and skip the optional rewrite. Why: Lance selects candidates by row count while the re-encoding writer splits output by bytes, so a task whose rows are too wide for the target emits outputs that are immediately re-selected - an unbounded rewrite loop (measured: a ~1 GB table rewrote ~30 GB across 31 syncs with zero net progress). A task whose whole volume fits one output cannot strand sub-target outputs, so the width test applies only beyond that point.
+
 ### 3.5 Concurrency
 
 pond processes are stateless workers. Several may write the same namespace at once; Lance optimistic concurrency control resolves append conflicts through manifest versioning. There is no external coordinator - object stores provide atomic conditional writes, the local filesystem uses Lance's commit lock - and no in-process write queue.
@@ -215,8 +219,6 @@ Writes commit data without folding indexes; index maintenance is operator-trigge
 | IVF_SQ (vector) | `optimize_indices(append)` | Stable-row-id IVF supports incremental fold via `IvfIndexBuilder::new_incremental`; centroids and per-dimension SQ ranges carry forward. |
 
 Index maintenance skips indexes whose columns no write has touched; this is sound because Lance prunes index coverage only for indexes whose fields overlap a write's modified fields. Why operator-triggered: matching Lance's own design (`table.optimize()` is the canonical periodic-maintenance op) avoids per-write index rebuilds that would otherwise dominate write cost.
-
-The default compaction filter also requires a planned non-deletion task's row target to fit within half the configured output byte budget at every input fragment's observed live-row width. The headroom covers the re-encoding writer's soft byte limit and binary-copy compaction's whole-file granularity. Missing file sizes skip the optional rewrite. Deletion materialization remains eligible because removing tombstoned rows is useful even when fragment layout does not improve.
 
 ### 3.8 Forward-compatibility seams
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -216,6 +216,8 @@ Writes commit data without folding indexes; index maintenance is operator-trigge
 
 Index maintenance skips indexes whose columns no write has touched; this is sound because Lance prunes index coverage only for indexes whose fields overlap a write's modified fields. Why operator-triggered: matching Lance's own design (`table.optimize()` is the canonical periodic-maintenance op) avoids per-write index rebuilds that would otherwise dominate write cost.
 
+The default compaction filter also requires a planned non-deletion task's row target to fit within half the configured output byte budget at every input fragment's observed live-row width. The headroom covers the re-encoding writer's soft byte limit and binary-copy compaction's whole-file granularity. Missing file sizes skip the optional rewrite. Deletion materialization remains eligible because removing tombstoned rows is useful even when fragment layout does not improve.
+
 ### 3.8 Forward-compatibility seams
 
 #### `lance-forward-compat`
--- a/packages/pond/src/substrate.rs
+++ b/packages/pond/src/substrate.rs
@@ -5231,15 +5231,6 @@ mod tests {
         // Remainder reaches largest / COMPACTION_ABSORB_FACTOR -> kept.
         let tiered = [stat(400_000), stat(60_000), stat(40_000)];
         assert!(task_is_kept(&tiered, derived_target_rows(&tiered)));
-
-        let off_boundary = [
-            fragment(99_900_000, 100_000, 0),
-            fragment(100_100_000, 100_000, 0),
-        ];
-        assert!(task_is_kept(
-            &off_boundary,
-            derived_target_rows(&off_boundary),
-        ));
     }
 
     #[test]
@@ -5279,7 +5270,24 @@ mod tests {
     fn compaction_veto_uses_physical_rows_after_deletions() {
         let partially_deleted = || fragment(100_000_000, 100_000, 9_000);
         let task: Vec<FragmentStat> = std::iter::repeat_with(partially_deleted).take(3).collect();
-        assert!(task_is_kept(&task, 134_217));
+        assert!(task_is_kept(&task, derived_target_rows(&task)));
+    }
+
+    #[test]
+    fn compaction_filter_keeps_above_budget_off_boundary_task() {
+        let table = [
+            fragment(100_000_000, 150_000, 0),
+            fragment(100_000_000, 150_000, 0),
+            fragment(100_000_000, 150_000, 0),
+            fragment(100_000_000, 50_000, 0),
+        ];
+        let task = &table[..3];
+        let target = derived_target_rows(&table);
+        let total_bytes = task.iter().map(|stat| stat.bytes.unwrap()).sum::<u64>();
+
+        assert!(total_bytes > TARGET_FRAGMENT_BYTES);
+        assert!(target < derived_target_rows(task));
+        assert!(task_is_kept(task, target));
     }
 
     #[test]
--- a/packages/pond/src/substrate.rs
+++ b/packages/pond/src/substrate.rs
@@ -833,9 +833,9 @@ fn classify_check_error(
     }
 }
 
-/// Per-task fragment-count backstop: tasks this wide always run, bounding
-/// manifest growth even when the amplification veto would skip them. As
-/// policy cap, 0 disables the veto (tests).
+/// Per-task fragment-count backstop: tasks this wide bypass the amplification
+/// veto once their row target is attainable, bounding manifest growth. As
+/// policy cap, 0 disables all task filtering (tests).
 pub const DEFAULT_COMPACTION_FRAGMENT_CAP: usize = 64;
 
 /// Fragments are sized by bytes, not Lance's 1M-row default: kilobyte-average
@@ -843,7 +843,6 @@ pub const DEFAULT_COMPACTION_FRAGMENT_CAP: usize = 64;
 /// re-rewrites wholesale to absorb tiny appends (~190 GiB/day of churn).
 pub const TARGET_FRAGMENT_BYTES: u64 = 256 * 1024 * 1024;
 
-const MIN_TARGET_ROWS_PER_FRAGMENT: u64 = 50_000;
 /// Ceiling = Lance's own default.
 const MAX_TARGET_ROWS_PER_FRAGMENT: u64 = 1024 * 1024;
 
@@ -1003,16 +1002,16 @@ fn fragment_stat(fragment: &lance::table::format::Fragment) -> FragmentStat {
     }
 }
 
-/// Candidacy/merge target: HALF the rows a [`TARGET_FRAGMENT_BYTES`] fragment
-/// holds at the table's average row size. Compaction byte-caps every output
-/// fragment at [`TARGET_FRAGMENT_BYTES`] (`max_bytes_per_file`), so deriving the
-/// target at the FULL byte budget made `target == the largest fragment
-/// compaction can produce`: no output could ever satisfy `physical_rows >=
-/// target`, so the table was re-compacted every sync for a net-zero fragment
-/// change (measured ~100-120s/sync on the remote store, 30->30 fragments).
+/// Candidacy/merge target: HALF the rows the [`TARGET_FRAGMENT_BYTES`] output
+/// budget holds at the table's average row size. Deriving the target at the
+/// FULL byte budget made `target == the largest fragment re-encoding could
+/// reliably produce`: no output could ever satisfy `physical_rows >= target`,
+/// so the table was re-compacted every sync for a net-zero fragment change
+/// (measured ~100-120s/sync on the remote store, 30->30 fragments).
 /// Halving leaves 2x headroom so a byte-capped fragment lands comfortably above
 /// the target and FREEZES, making compaction productive (merge small -> freeze
-/// -> stop) instead of perpetual churn.
+/// -> stop) instead of perpetual churn. A row floor would make the target
+/// unreachable again for sufficiently wide rows.
 fn derived_target_rows(stats: &[FragmentStat]) -> usize {
     let (mut bytes, mut rows) = (0u64, 0u64);
     for stat in stats {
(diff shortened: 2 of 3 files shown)


alexnayko:
CI is waiting for maintainer approval on run 30107696259 because this is a fork contribution. Could you approve the workflow when convenient? I will address any failures here.

tenequm:
Thanks for the contribution - CI is approved and running now. I'll give this a thorough review.

packages/pond/src/substrate.rs
```diff
@@ -1026,32 +1025,55 @@ fn derived_target_rows(stats: &[FragmentStat]) -> usize {
...
         stat.rows > 0 && (stat.deleted_rows as f32 / stat.rows as f32) > deletion_threshold
     }) {
         return true;
     }
+    let mut total_bytes = 0u128;
+    for stat in stats {
+        let Some(bytes) = stat.bytes.map(u128::from) else {
+            return false;
+        };
+        let live_rows = u128::from(stat.rows.saturating_sub(stat.deleted_rows));
```
tenequm: The rewrite output writes only live rows and only their bytes, so the correct predictor of output width is `bytes / physical_rows`; dividing by live rows inflates apparent width by 1/(1-d) for deletion-bearing fragments - exactly the ones a future purge path needs rewritten. Please use `stat.rows` here (the sketch in the review summary already does). `compaction_veto_uses_live_rows_after_deletions` currently locks in the inverted behavior and will need to flip.

packages/pond/src/substrate.rs
```diff
@@ -2982,13 +3004,15 @@ async fn optimize_table_compact(
                 &task_stats,
                 policy.compaction_fragment_cap,
                 compaction.materialize_deletions_threshold,
+                compaction.target_rows_per_fragment,
+                TARGET_FRAGMENT_BYTES,
```
tenequm: Please pass `compaction.max_bytes_per_file` (unwrapped) instead of re-passing the constant - `CompactionOptions` above holds the same number, and two sources for one value means the gate can silently diverge from what the writer was actually told.

packages/pond/src/substrate.rs
```diff
@@ -2982,13 +3004,15 @@ async fn optimize_table_compact(
...
             );
             if !keep {
                 tracing::debug!(
                     target: "pond::perf",
                     table = table.as_str(),
                     fragments = task_stats.len(),
-                    "compaction task vetoed: merge dominated by one large fragment",
+                    "compaction task vetoed: unreachable row target or excessive amplification",
```
tenequm: A permanently vetoed table is otherwise invisible - the phase still reports Ok. Please add a `reason` field distinguishing at least: missing sizes / cannot shrink / row target unattainable / absorb veto.

packages/pond/src/substrate.rs
```diff
@@ -5162,49 +5186,91 @@ mod tests {
...
+
+    #[test]
+    fn compaction_veto_uses_live_rows_after_deletions() {
+        let partially_deleted = || fragment(100_000_000, 100_000, 9_000);
+        let task: Vec<FragmentStat> = std::iter::repeat_with(partially_deleted).take(2).collect();
+        assert!(!task_is_kept(&task, 134_217));
+    }
+
+    #[test]
+    fn compaction_veto_rejects_byte_capped_second_cycle() {
```
tenequm, line 5318: Two coverage asks:

- The kept-case fixtures sit on exact boundary equality (`stat()` synthesizes width 1,000 and `134_217 = floor(128MiB/1000)`), so a change to `TARGET_FRAGMENT_BYTES` flips them silently - please add one off-boundary case, and prefer `derived_target_rows(&fixture)` over the bare literal so the coupling is visible.
- Please add a mixed-width KEPT case shaped like a real planned task - this is the shape that catches the regression class described in the review summary. From my live manifests, `(rows, bytes)` with `target = 58_468`: `[(9_281, 26_612_870), (4_400, 14_242_314), (20_988, 54_111_122), (166, 517_923)]` (total under one 256 MiB output - must be kept), and the 9-fragment task `[(3_946, 13_547_709), (155, 344_320), (34, 134_209), (15_759, 54_624_719), (292, 1_364_162), (32_826, 110_225_801), (1_840, 8_151_118), (79, 728_590), (128, 685_184)]`.

The fail-closed and deletion-precedence tests are good - keep them (fail-closed is fine as a policy: I verified Lance >= 0.27.0 backfills `file_size_bytes` at commit, so it's unreachable for pond-written stores anyway).

docs/spec.md
```diff
@@ -216,6 +216,8 @@ Writes commit data without folding indexes; index maintenance is operator-trigge
 
 Index maintenance skips indexes whose columns no write has touched; this is sound because Lance prunes index coverage only for indexes whose fields overlap a write's modified fields. Why operator-triggered: matching Lance's own design (`table.optimize()` is the canonical periodic-maintenance op) avoids per-write index rebuilds that would otherwise dominate write cost.
 
+The default compaction filter also requires a planned non-deletion task's row target to fit within half the configured output byte budget at every input fragment's observed live-row width. The headroom covers the re-encoding writer's soft byte limit and binary-copy compaction's whole-file granularity. Missing file sizes skip the optional rewrite. Deletion materialization remains eligible because removing tombstoned rows is useful even when fragment layout does not improve.
```
tenequm: Two issues here: placement (3.7 is index lifecycle; fragment-sizing policy belongs in 3.4 Dataset parameters), and the WHY doesn't survive contact with Lance v8 - the reencode soft cap overshoots (favorable direction), and binary copy ignores `max_bytes_per_file` outright (binary_copy.rs:247), so neither justifies the headroom. Please replace this paragraph with a named rule in section 3.4, after `lance-table-creation`:

```markdown
#### `lance-compaction-filter`

Automatic compaction runs only Lance-planned tasks that pass pond's filter. Tasks above the deletion-materialization threshold always run - reclaiming tombstones is useful even when layout does not improve. Any other task must strictly reduce the fragment count, and when its total bytes exceed one output file's byte budget, every input fragment's observed row width must let an output reach the row target within half that budget (headroom because re-encoding can widen rows relative to the input estimate). Missing file sizes fail closed and skip the optional rewrite. Why: Lance selects candidates by row count while the re-encoding writer splits output by bytes, so a task whose rows are too wide for the target emits outputs that are immediately re-selected - an unbounded rewrite loop (measured: a ~1 GB table rewrote ~30 GB across 31 syncs with zero net progress). A task whose whole volume fits one output cannot strand sub-target outputs, so the width test applies only beyond that point.
```

tenequm (changes requested):
Thank you - this is an exceptional report and PR. I reviewed it deeply: replayed Lance v8.0.0's planner over my real store manifests, simulated both policies over multi-hundred-sync horizons, and ran your branch locally - it is fully green (fmt, clippy -D warnings, 362/362 tests). I reproduced your incident in simulation from the floor alone (31 rounds x ~980 MiB, net zero) and confirmed it converges in one pass once the floor is gone. The floor removal, the u128 target derivation, and filtering whole planned tasks (explicitly sanctioned by upstream: optimize.rs:78-82) are all right and I want this merged.

One structural amendment is needed on the gate; the smaller items are inline on the diff so everything can land in a single revision.

## Condition the gate on the byte cap actually being able to split the task

As written, the gate reduces to "veto any task containing a fragment wider than the table's average row width": since `target = floor(128MiB * rows / bytes)`, the per-fragment condition `bytes * target * 2 <= rows * 256MiB` is equivalent to `width_f <= avg_width` with near-zero tolerance. On real corpora the median fragment sits above the mean (small fragments compress worse; a few large well-compressed fragments drag the average down), so on my store this vetoes every currently planned task on all three tables, and simulating 300 syncs of ordinary appends grows one table from 20 to 320 fragments with zero rewrites. The 2x headroom is also applied twice - `derived_target_rows` already halves the budget so a byte-capped output lands at 2x the target.

Two Lance v8.0.0 facts sharpen where the gate is actually needed:

1. Reencode outputs split at `min(target_rows, max_bytes)` - the target is passed through as `max_rows_per_file` (optimize.rs:1615) - and the byte cap only overshoots (checked after each batch on flushed bytes, write.rs:646-648), which adds rows and helps freezing.
2. The binary-copy path never reads `max_bytes_per_file` at all (binary_copy.rs:247; it flushes on the row target at whole-input-file granularity), so binary-copy-eligible tasks cannot loop regardless of width.

So the loop exists only when the byte cap can split a task's output. A task whose total bytes fit one output file merges to a single fragment, which is then a lone sub-target candidate that the planner drops as a noop (optimize.rs:1353-1364) - it terminates unconditionally. Gating on that precondition, plus requiring the merge to strictly shrink the fragment count (which restores the property the 64-fragment backstop existed for), keeps every convergence property you wanted while preserving current behavior everywhere it was correct:

```rust
let mut total_bytes = 0u128;
let mut largest = 0u128;
for stat in stats {
    let Some(bytes) = stat.bytes.map(u128::from) else {
        return false;
    };
    total_bytes += bytes;
    largest = largest.max(bytes);
}
let budget = u128::from(max_bytes_per_file);
// A rewrite that cannot end with fewer fragments than it started is pure
// write amplification regardless of widths.
if u128::try_from(stats.len()).unwrap_or(u128::MAX) <= total_bytes.div_ceil(budget).max(1) {
    return false;
}
if stats.len() >= cap {
    return true;
}
// Only a byte-split rewrite can strand outputs below the row target; a task
// whose whole volume fits one output merges to a single fragment and stops.
if total_bytes > budget {
    for stat in stats {
        let bytes = stat.bytes.map(u128::from).unwrap_or(0);
        let rows = u128::from(stat.rows);
        if rows == 0 || bytes * target_rows_per_fragment as u128 * 2 > rows * budget {
            return false;
        }
    }
}
(total_bytes - largest) * u128::from(COMPACTION_ABSORB_FACTOR) >= largest
```

I simulated exactly this rule against both current main and your PR: it matches main on my real tables (messages 20 -> 14 fragments over 300 syncs), matches your PR on the incident shape (one full pass, then silence), and also kills a residual loop that survives floor removal alone (a contiguous run of locally-wide fragments in a narrow-average table: measured 55.8 GiB over 31 simulated rounds under current main - your gate instinct was right, it just needs this precondition).

Again - genuinely impressive work, both the forensics and the restraint of the diff. With these in, I'll merge promptly.

packages/pond/src/substrate.rs
alexnayko: Fixed in [`b7a1727`](https://github.com/alexnayko/pond/commit/b7a1727b8730f36dcdb40489cb20a0f87deb5b50): the width check now uses `stat.rows`, and `compaction_veto_uses_physical_rows_after_deletions` now asserts that the deletion-bearing task is kept.

packages/pond/src/substrate.rs
alexnayko: Fixed in [`b7a1727`](https://github.com/alexnayko/pond/commit/b7a1727b8730f36dcdb40489cb20a0f87deb5b50): the filter now receives `compaction.max_bytes_per_file`. Its `Option<usize>` conversion is checked and fails closed as `invalid_byte_budget` when absent or out of range.

packages/pond/src/substrate.rs
alexnayko: Fixed in [`b7a1727`](https://github.com/alexnayko/pond/commit/b7a1727b8730f36dcdb40489cb20a0f87deb5b50): `task_veto_reason` now logs `missing_sizes`, `cannot_shrink`, `row_target_unattainable`, `absorb_veto`, or `invalid_byte_budget`.

packages/pond/src/substrate.rs
alexnayko: Fixed in [`b7a1727`](https://github.com/alexnayko/pond/commit/b7a1727b8730f36dcdb40489cb20a0f87deb5b50): the coupled fixtures now use `derived_target_rows`, and the suite includes an off-boundary case plus both real 4-fragment and 9-fragment mixed-width tasks from your comment.

docs/spec.md
alexnayko: Fixed in [`b7a1727`](https://github.com/alexnayko/pond/commit/b7a1727b8730f36dcdb40489cb20a0f87deb5b50): the named `lance-compaction-filter` rule now sits in section 3.4 with the requested contract and rationale.

alexnayko:
All five inline comments are addressed in [`b7a1727`](https://github.com/alexnayko/pond/commit/b7a1727b8730f36dcdb40489cb20a0f87deb5b50) and replied to individually.

Local validation is green: `cargo fmt`, `cargo clippy --locked --offline -- -D warnings`, and 363 tests passed with 1 ignored.

The real-store canary also converged. The first normal `pond optimize --only index` reduced active `parts` fragments from 14 to 6. An immediate second pass made zero file changes. Row counts, corpus hashes, protected transcript hashes, and the FTS result remained identical.

The new [CI run](https://github.com/tenequm/pond/actions/runs/30167918970) is waiting for external-fork approval. Could you approve it when convenient?

packages/pond/src/substrate.rs
alexnayko: Correction to my earlier reply: `b7a1727` still left one bare target literal, and its 200 MB off-boundary fixture stayed below the 256 MiB budget, so it did not enter the conditional width branch. [`e74de1f`](https://github.com/alexnayko/pond/commit/e74de1fb12fff2decaa85087d7ac6ba3af556049) now derives the deletion fixture target and replaces that case with a 300 MB planned task. Its target comes from the wider whole-table fixture and differs from the task-local target, so the test exercises and keeps a genuinely off-boundary task above budget.

tenequm (approved):
All five points addressed exactly as asked, and improved on in places - the Option<&'static str> reason design is cleaner than what I sketched, and the restored 64-fragment backstop now terminates via cannot_shrink on the following cycle instead of looping, which I verified by hand against the new fixtures. Ran your branch locally: fmt and clippy clean, 364 tests passing. The real-store canary result (14 -> 6 fragments, second pass zero changes) matches what my simulations predicted. Merging once CI is green. Thank you for a genuinely excellent contribution - the report, the fix, and the revision discipline were all first-rate.

tenequm:
Released in [v0.14.3](https://github.com/tenequm/pond/releases/tag/v0.14.3) - crates.io, Homebrew, and the flake all carry it now. The release notes credit you and this investigation directly. Thanks again for a first-rate contribution; the follow-up on #128/#129 is being reviewed as well.

PR #129, opened 2026-07-26
fix(cli): remove unused tracing progress layer

### LINKED ISSUE #128: Unfiltered IndicatifLayer lets pond -vv hit an upstream non-TTY panic
## Summary

`pond optimize -vv` can panic when stdout and stderr are combined and piped through `tee`.

This was observed while testing #124. Two normal non-verbose production optimize runs completed, so this report is separate from the compaction behavior in that PR.

## Environment

- macOS on aarch64
- affected Pond code path is present at PR #124 head `e74de1fb12fff2decaa85087d7ac6ba3af556049` (`pond-db` 0.14.2)
- `tracing-indicatif` 0.3.14
- `indicatif` 0.18.4
- `bytes` 1.11.1
- `tracing` 0.1.44
- `tracing-subscriber` 0.3.23

## Command

The failure occurred with the optimize output captured this way:

```bash
set -o pipefail
sandbox_root=/private/tmp/pond-verbose-panic-repro
env \
  HOME="$sandbox_root/home" \
  XDG_CONFIG_HOME="$sandbox_root/xdg/config" \
  XDG_DATA_HOME="$sandbox_root/xdg/data" \
  XDG_STATE_HOME="$sandbox_root/xdg/state" \
  XDG_CACHE_HOME="$sandbox_root/xdg/cache" \
  POND_STORAGE_PATH="$sandbox_root/store" \
  POND_CONFIG_FILE="$sandbox_root/xdg/config/pond/config.toml" \
  RUST_BACKTRACE=full \
  pond optimize -vv \
    --storage-path "$sandbox_root/store" \
    --config-file "$sandbox_root/xdg/config/pond/config.toml" \
    2>&1 | tee "$sandbox_root/optimize.log"
```

The reproduction used an isolated HOME, XDG tree, config, and Pond store. It did not enable scheduling or sync against the default local store.

## Actual behavior

The first panic was:

```text
tracing-indicatif-0.3.14/src/pb_manager.rs:160
footer progress bar was hidden despite there being pending progress bars
```

It was followed by:

```text
bytes-1.11.1/src/bytes.rs:392
range end out of bounds: 91504 <= 0
```

## Expected behavior

Verbose optimize should finish or return a normal error. Redirecting output through `tee` should not cause a panic.

## Reduced control

The same command completed with exit 0 on installed Pond 0.14.1 (`279856a`, aarch64-macos) against a fresh isolated fixture store containing 3 sessions, 576 total messages, and 24 searchable messages. `pond status -vv` with the same combined pipe also exited 0. This indicates that the pipe and verbosity expose a workload-sensitive span-concurrency bug.

## Source diagnosis

Pond installs `IndicatifLayer::new()` without a per-layer filter:

https://github.com/tenequm/pond/blob/b6426140d97ef4a3f6a4d186252e1028253cabdc/packages/pond/src/main.rs#L1652-L1664

`tracing-indicatif` documents that the layer performs no filtering and expects callers to attach a filter when progress bars should be opt-in:

https://github.com/emersonford/tracing-indicatif/blob/0.3.14/src/lib.rs#L347-L354

At `-vv`, Pond therefore turns enabled Lance and DataFusion spans into progress bars. Once more than the default seven overlap, `tracing-indicatif` uses its pending-footer path. Its open issue #24 reports this exact assertion without a terminal: `ProgressBar::is_hidden()` is true for non-terminal bars as well as queued bars, which corrupts the pending-bar accounting.

https://github.com/emersonford/tracing-indicatif/issues/24

The `tee` interaction is therefore relevant because `2>&1` makes the progress target non-terminal. The later `bytes` panic appears to be secondary during concurrent panic or output unwinding, but that part is not yet reduced independently.

## Focused containment in Pond

Pond currently has no `indicatif.pb_show` spans. Applying `IndicatifFilter::new(false)` to the `IndicatifLayer` would match the existing opt-in intent in the source comment and prevent unrelated dependency spans from entering the progress-bar manager. A regression should exercise more than seven spans with a non-terminal output target.

Related: #124

alexnayko:
## What changed

- remove the unused `tracing-indicatif` layer and dependency
- write tracing output directly through `tracing-subscriber`

## Why

Pond has no `indicatif.pb_show` spans. The unfiltered layer turned `-vv` dependency spans into progress bars and exposed the non-terminal panic tracked in `tracing-indicatif` issue #24:

https://github.com/emersonford/tracing-indicatif/issues/24

Its writer only suspends bars owned by its own `MultiProgress`. Pond's real progress bars are separate, so the layer did not protect them. Removing it eliminates the failing path and dead future plumbing.

## Checks

- `rustfmt 1.9.0-stable --edition 2024 --check packages/pond/src/main.rs`
- `cargo metadata --locked --offline --no-deps --format-version 1`
- `git diff --check`

A full compile remains with CI because this checkout has no reusable Cargo target and a prior cold Pond build used about 17.8 GiB.

Closes #128

### DIFF
--- a/packages/pond/src/main.rs
+++ b/packages/pond/src/main.rs
@@ -1648,9 +1648,10 @@ fn init_tracing(cli_level: tracing::level_filters::LevelFilter) {
             "{cli_level},lance::index::vector::builder=error,aws_config=error,lance::object_store::throttle=error,lance::io::exec::count_pushdown=error"
         ))
     });
+    // stdout is reserved for JSON-RPC frames and machine-readable command output.
     tracing_subscriber::registry()
         .with(filter)
-        .with(fmt::layer())
+        .with(fmt::layer().with_writer(std::io::stderr))
         .init();
 }
 
--- a/packages/pond/src/main.rs
+++ b/packages/pond/src/main.rs
@@ -123,8 +123,7 @@ impl From<CliSessionFrom> for SessionFrom {
 }
 use serde_json::{Value, json};
 use tokio::io::AsyncWriteExt;
-use tracing_indicatif::{IndicatifLayer, filter::IndicatifFilter};
-use tracing_subscriber::layer::{Layer, SubscriberExt};
+use tracing_subscriber::layer::SubscriberExt;
 use tracing_subscriber::util::SubscriberInitExt;
 use tracing_subscriber::{EnvFilter, fmt};
 use url::Url;
@@ -1626,15 +1625,6 @@ async fn main() -> anyhow::Result<()> {
     Ok(())
 }
 
-fn opt_in_indicatif_layer<S>(
-    layer: IndicatifLayer<S>,
-) -> tracing_subscriber::filter::Filtered<IndicatifLayer<S>, IndicatifFilter<S>, S>
-where
-    S: tracing::Subscriber + for<'a> tracing_subscriber::registry::LookupSpan<'a>,
-{
-    layer.with_filter(IndicatifFilter::new(false))
-}
-
 fn init_tracing(cli_level: tracing::level_filters::LevelFilter) {
     // Lance's IVF_SQ builder warns once per empty centroid during merge
     // (rust/lance/src/index/vector/builder.rs: "partition N is empty, skipping").
@@ -1658,16 +1648,9 @@ fn init_tracing(cli_level: tracing::level_filters::LevelFilter) {
             "{cli_level},lance::index::vector::builder=error,aws_config=error,lance::object_store::throttle=error,lance::io::exec::count_pushdown=error"
         ))
     });
-    // Keep progress bars opt-in. Without the filter, `-vv` turns every Lance
-    // and DataFusion span into a bar, and tracing-indicatif can panic when its
-    // pending footer is hidden by a non-terminal stderr target.
-    let indicatif_layer = IndicatifLayer::new();
-    let fmt_layer = fmt::layer().with_writer(indicatif_layer.get_stderr_writer());
-    let indicatif_layer = opt_in_indicatif_layer(indicatif_layer);
     tracing_subscriber::registry()
         .with(filter)
-        .with(fmt_layer)
-        .with(indicatif_layer)
+        .with(fmt::layer())
         .init();
 }
 
@@ -5953,22 +5936,6 @@ mod tests {
     #![allow(clippy::expect_used, clippy::unwrap_used)]
 
     use super::*;
-
-    #[test]
-    fn indicatif_layer_ignores_more_than_seven_unmarked_spans() {
-        fn enter_unmarked_spans(depth: usize) {
-            if depth == 0 {
-                return;
-            }
-            let span = tracing::info_span!("unmarked_test_span", depth);
-            let _guard = span.enter();
-            enter_unmarked_spans(depth - 1);
-        }
-
-        let subscriber =
-            tracing_subscriber::registry().with(opt_in_indicatif_layer(IndicatifLayer::new()));
-        tracing::subscriber::with_default(subscriber, || enter_unmarked_spans(8));
-    }
 
     #[test]
     fn serve_bootstrap_acts_only_on_a_fully_unconfigured_pond() -> anyhow::Result<()> {
--- a/packages/pond/src/main.rs
+++ b/packages/pond/src/main.rs
@@ -123,8 +123,8 @@ impl From<CliSessionFrom> for SessionFrom {
 }
 use serde_json::{Value, json};
 use tokio::io::AsyncWriteExt;
-use tracing_indicatif::IndicatifLayer;
-use tracing_subscriber::layer::SubscriberExt;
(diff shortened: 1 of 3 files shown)


alexnayko:
Could a maintainer please approve [CI run 30202613868](https://github.com/tenequm/pond/actions/runs/30202613868)? GitHub marked the fork-triggered workflow `action_required`, so no jobs have started yet.

tenequm:
Thanks for the PR, looks like a good cleanup on the first glance.
Will do a few checks before confirming, will post a review shortly.

tenequm (changes requested):
Removing the layer is the right call, and it does eliminate the #128 panic at the root - I verified `IndicatifWriter` only suspends the layer's own `MultiProgress` ([tracing-indicatif 0.3.14 `lib.rs:457`](https://github.com/emersonford/tracing-indicatif/blob/0.3.14/src/lib.rs#L457)), while every real pond bar is a standalone `ProgressBar`/`MultiProgress` on `ProgressDrawTarget::stderr_*` (`main.rs:4180`, `4267`, `5084`). The layer never protected them.

One blocker: the replacement silently moves tracing output from stderr to **stdout**, which corrupts the MCP JSON-RPC channel. Details inline on `main.rs:1653`.

**Checks run:** `cargo fmt --check` OK, `cargo clippy --all-targets -- -D warnings` clean. No orphaned `tracing_indicatif` references anywhere in the tree, `indicatif` correctly retained as a direct dep, `Cargo.lock` consistent, no stale docs.

**2 findings (1 critical, 1 minor) - both inline.**

packages/pond/src/main.rs
```diff
@@ -1649,18 +1648,9 @@ fn init_tracing(cli_level: tracing::level_filters::LevelFilter) {
...
-    // bar, so the layer is essentially a "safe stderr writer for fmt that
-    // coexists with future bar-rendering spans"; the migration cost when we
-    // do start instrumenting spans (e.g. ingest, optimize) is zero.
-    let indicatif_layer = IndicatifLayer::new();
-    let fmt_layer = fmt::layer().with_writer(indicatif_layer.get_stderr_writer());
     tracing_subscriber::registry()
         .with(filter)
-        .with(fmt_layer)
-        .with(indicatif_layer)
+        .with(fmt::layer())
```
tenequm: **Critical** - `fmt::layer()` defaults to **stdout**, so this moves all tracing output off stderr.

`tracing_subscriber::fmt::layer()` is `Layer::default()`, whose writer parameter defaults to `W = fn() -> io::Stdout`. The `.with_writer(indicatif_layer.get_stderr_writer())` removed just above was what put logs on stderr - dropping the layer dropped the writer with it.

Measured on this branch vs a `main` baseline, same sandbox, `RUST_LOG=info pond mcp` fed a single `initialize` frame:

| | non-JSON-RPC lines on stdout | stderr lines |
|---|---|---|
| `main` | 0 | 8 |
| this PR | **9** | 0 |

The 9 log lines interleave directly around the `initialize` response, so the JSON-RPC frame stream is corrupted.

This breaks four documented invariants:

- `packages/pond/src/transport.rs:1218` - "All diagnostics go to stderr (the shared `tracing` subscriber); stdout carries only JSON-RPC frames (spec.md#scope)"
- `packages/pond/src/main.rs:1375` - `eprintln!("serve: stdio MCP ready; stdout is reserved for JSON-RPC")`
- `docs/spec.md:695` - `pond serve` "routes all its output to stderr/tracing so stdout stays reserved for the transport"
- `docs/spec.md:688` - `pond sync --format json` "emits one machine-readable summary document on stdout ... (progress stays on stderr)" - a stray WARN now lands inside that document

This is not `-vv`-only: default verbosity is `WarnLevel` (`main.rs:290`), so any Lance / `aws_config` WARN that survives the `EnvFilter` hits stdout with no flags at all. Second data point: `pond status -vv` in a sandbox put 281 log lines on stdout mixed into the status report, and 0 on stderr.

```suggestion
        .with(fmt::layer().with_writer(std::io::stderr))
```

tenequm: **Minor** - the stderr choice deserves a WHY comment.

`init_tracing` is already densely commented with load-bearing WHYs (`main.rs:1629-1645` explains each `EnvFilter` directive). Once the writer is restored, stderr becomes the only thing keeping the MCP JSON-RPC channel and the `--format json` surfaces clean - and its implicitness is exactly what this change tripped over. A one-liner would stop the next refactor from dropping it again.

e.g. `// stderr is load-bearing: stdout carries JSON-RPC frames and --format json documents (spec.md#scope).`

alexnayko: Confirmed independently in `tracing-subscriber` 0.3.23: the default fmt writer is stdout. Fixed in 5c37229 with `.with_writer(std::io::stderr)`, preserving stdout for JSON-RPC frames and machine-readable command output.

alexnayko: Added the load-bearing WHY comment in 5c37229 next to the explicit stderr writer: stdout is reserved for JSON-RPC frames and machine-readable command output.

PR #147, opened 2026-08-12
feat: Windows support

BadAd84:
## Summary

This brings pond up on **Windows** (`x86_64-pc-windows-msvc`) - build, full test suite, `clippy -D warnings`, and `fmt` all green - without regressing Linux or macOS. Every change is genuinely cross-platform (a shared code path or a per-OS branch), never a Windows-only fork bolted alongside the existing behaviour.

Verified end to end by syncing **331 oh-my-pi sessions** on Windows 11 (embed on the CPU backend, Lance datasets committing to `C:\...` via `file:///C:/...`), plus `pond status`, `pond schedule`, and the fmt/clippy/test gates.

## Scope note for the maintainer

`docs/spec.md` §2.4 currently reads "Linux and macOS. Windows is not in v1 scope." This PR delivers Windows and updates §2.4 and the README accordingly, but whether Windows belongs in v1 is your call - happy to drop the spec/README/CI parts and keep this as an out-of-tree patch, or adjust the framing, if you'd prefer. Raising it explicitly rather than quietly changing the stated scope.

If this is larger than you want to review in one go, I'm happy to split it into stages and submit them in sequence for easier review - a natural breakdown would be: (1) build enablement (the `protoc` feature split + the dedicated-thread `main` startup fix + `home_dir()`), the minimal "it builds and runs" core; (2) the Task Scheduler backend; (3) the `*_command` shell and sync-lock holder fixes; (4) the cross-platform test adjustments and the docs/spec updates. Each stage stands alone and keeps the suite green. Just say which shape you'd prefer.

## What changed

**Build**
- lance's `protoc` feature vendors protoc via `protobuf-src` (built from source), which does not link with MSVC (absl/UCRT unresolved externals). The feature is made unix-only in both `[dependencies]` and `[dev-dependencies]`; Windows uses a system `protoc` on `PATH`. Unix keeps vendoring exactly as before.

**Runtime**
- `main` ran the entire async CLI future on the main thread. Windows' default main-thread stack is 1 MiB (vs 8 MiB on Unix), so the large state machine overflowed at startup - even `--version`. It now runs the runtime on a dedicated 16 MiB thread; identical behaviour on every platform.
- Home resolution read only `$HOME`, which Windows does not set - this silently killed all adapter auto-discovery. Added a per-OS `config::home_dir()` (`USERPROFILE` on Windows, `HOME` elsewhere) and routed every home-relative default path and discovery probe through it. It is decided by OS, not a `HOME`-then-`USERPROFILE` fallback, deliberately: a native binary launched from Git Bash inherits a POSIX-style `$HOME` that must not be followed.
- `pond schedule` gains a **Task Scheduler** backend (`schtasks`), the peer of the launchd/systemd/cron `unix` module: `start`/`stop`/`status`/`logs`, indefinite repeat, exit codes, and cadence recovered from the task XML. The scheduled action is a generated `.cmd` wrapper (the Windows analogue of the plist/cron artifact), which keeps pond out of `schtasks` command-line quoting.
- `*_command` secrets (spec `storage-redaction`) ran `sh -c`; they now use the platform shell (`cmd /C` on Windows).
- The sync single-flight lock stored its holder breadcrumb inside the flock'd file. Windows takes a mandatory whole-file lock, so a blocked sibling could not read the holder out of the locked file. The breadcrumb moved to a sibling unlocked file, readable on both platforms.

**Tests** (all made cross-platform, still exercising the real behaviour)
- Path/URL translation tests use per-OS absolute paths (a driveless `/xdg` is not absolute on Windows), so they now validate Windows drive-path round-trips too.
- The secret-`_command` test uses per-OS shell commands.
- Path-layout assertions compare on `/`-normalized text; the on-disk paths keep OS separators.
- The rowmap corrupt-and-rebuild test drops the store's `mmap` before overwriting the segment files (Windows forbids writing a file with a live mapping; POSIX allows it). Production cleanup was already best-effort here, so only the test needed adjusting.

**Docs**
- README install section notes the Windows path (`cargo install` / build-from-source, needs `protoc`; Homebrew/Nix stay macOS/Linux). `docs/spec.md` §2.4 updated. (No CI change - per your note, the pipeline already builds/publishes the Windows binary.)

## Non-regression

No `#[cfg(windows)]` was introduced where a single cross-platform path works. The `protoc` split leaves Unix vendoring untouched; the dedicated-thread `main` is behaviourally identical on Unix (which already had ample stack); `home_dir()` reads `HOME` first-and-only on Unix. Full suite + clippy + fmt were re-run on this machine after every change.

Cargo.lock is unchanged (the feature move added/removed no locked packages).


tenequm:
Thanks for the submission, will review it thoroughly tomorrow!

tenequm:
@BadAd84 small notice, no need to add an additional ci job to the workflow.
Windows binary is already being built and published in the pipeline. The missing part should mostly be only configs.

packages/pond/Cargo.toml
```diff
@@ -103,7 +103,10 @@ fastrand = "2"
 figment = { version = "0.10.19", features = ["toml", "env"] }
 hf-hub = { version = "0.5", default-features = false, features = ["ureq"] }
 indicatif = "0.18"
-lance = { version = "8.0.0", features = ["protoc"] }
+# On unix, lance's `protoc` feature vendors protoc via protobuf-src (built from
+# source). protobuf-src does not link on Windows (absl/UCRT unresolved externals),
+# so Windows builds omit the feature and use a system protoc found via PATH/PROTOC.
+lance = { version = "8.0.0" }
```
tenequm, line 110: **[critical]** This gate breaks the existing Windows release build. `moon.yml` `build-dist` already cross-compiles `x86_64-pc-windows-gnu` via cargo-zigbuild on a Linux runner and publishes `pond-x86_64-pc-windows-gnu.exe` - and that build currently succeeds *because* the `protoc` feature vendors protoc via `protobuf-src`. Gating on `cfg(unix)` turns vendoring off for every Windows target, including windows-gnu, and the runner has no system `protoc` (nothing in `moon.yml` or the runner image installs one), so the next release fails at that target.

The linkage failure you hit is MSVC-specific (absl/UCRT unresolved externals), so scope the workaround to where it actually fails:

```toml
[target.'cfg(not(target_env = "msvc"))'.dependencies]
lance = { version = "8.0.0", features = ["protoc"] }
```

(and the same for the dev-dependencies split). Workspace resolver is `"3"`, so per-target feature resolution works correctly either way.

Minor aside: the `[target.'cfg(unix)'.dev-dependencies]` `lance` re-declaration at line 226 is redundant - on unix the feature already unifies in from the target dependency.

packages/pond/src/schedule.rs
```diff
@@ -914,3 +917,317 @@ mod tests {
...
+        let log = log_path();
+        let state = crate::syncstate::pond_state_dir();
+        std::fs::create_dir_all(&state)
+            .with_context(|| format!("failed to create {}", state.display()))?;
+
+        // The task runs this wrapper (.cmd files are executed through cmd, so
+        // the `>>` redirection works) instead of a quoted `schtasks /TR`
+        // command line, keeping every path out of schtasks quoting.
+        let wrapper = wrapper_path();
+        let wrapper_body = format!(
```
tenequm, line 1000: **[critical]** The wrapper never pins `XDG_STATE_HOME`, which `docs/spec.md` section 7.8 makes a contract: *"Registration pins the resolved `XDG_STATE_HOME` into the job's environment"*. All three unix backends do it (plist `EnvironmentVariables`, systemd `Environment=`, cron line prefix), with the WHY at the top of `start()`: Task Scheduler never sources shell config, so a user who sets `XDG_STATE_HOME` in their PowerShell `$PROFILE` (not `setx`) gets a scheduled sync that resolves a *different* state dir than manual syncs - splitting the single-flight flock, so a tick and a manual sync run concurrently against the same store, and `pond status` reads a last-sync record the scheduled run never writes.

Fix: bake `set "XDG_STATE_HOME=..."` into the wrapper. Careful which value: unix pins `state_root()`; this function binds `state = pond_state_dir()`, which is `state_root().join("pond")` - pinning that would make the child resolve `<root>/pond/pond`.

Two related wrapper issues while you are in here:
- `%` is expansion-active in batch files even inside double quotes, so a `%` in the binary or state path (legal in NTFS names and usernames) silently corrupts the command - the scheduled sync never runs while `pond status` still reports the schedule active. Double `%` to `%%`, or apply the same character gate `start()` uses for the unix templates (which rejects `%` for exactly this embedding reason).
- The wrapper is written before `/Create`, so a failed `/Create` leaves an orphaned `pond-sync.cmd`. Trivial: write it after, or note that stop() intentionally leaves artifacts behind anyway.

packages/pond/src/substrate.rs
```diff
@@ -608,8 +608,15 @@ fn run_secret_command(set: &str, field: &str, command: &str) -> Result<String> {
     {
         return Ok(hit.clone());
     }
-    let output = std::process::Command::new("sh")
-        .arg("-c")
+    // The command is a shell line (pipes, `op read`, etc.), so run it through
+    // the platform shell: `cmd /C` on Windows, `sh -c` elsewhere.
+    let (shell, shell_flag) = if cfg!(windows) {
```
tenequm: **[critical]** `Command::arg` applies MSVCRT escaping unconditionally on Windows (inner `"` become `\"`, the arg gets wrapped in quotes when it contains a space), and `cmd.exe` does not parse by MSVCRT rules - it has no backslash escape. Std's own docs for `CommandExt::raw_arg` name this exact case: it exists *"for passing arguments to applications that don't follow the standard C run-time escaping rules, such as cmd.exe /c"*. (The `.bat`/`.cmd` special case from CVE-2024-24576 does not apply here - it triggers only when the *program* has that extension, and the program here is cmd itself.)

Concrete failure: `secret_access_key_command = 'op read "op://My Vault/aws/secret"'` - quotes are mandatory the moment a vault or item name has a space. Rust emits `cmd /C "op read \"op://My Vault/aws/secret\""`; cmd sees four quote chars, strips the first and last, and `op` receives mangled arguments. The unix `sh -c` path handles the same string fine, and spec `storage-redaction` makes `_command` a first-class secret source, so this breaks S3 creds on Windows for any quoted command.

Fix: `std::os::windows::process::CommandExt::raw_arg` for the command string on the cmd branch. Minor aside: consider `%COMSPEC%` with a `cmd` fallback rather than resolving `cmd` through `PATH`.

The new test at line 5066 (`echo from-command&echo.`) contains no quotes, so it cannot catch this - please add a case whose command carries a space and a quote.

packages/pond/src/schedule.rs
```diff
@@ -914,3 +917,317 @@ mod tests {
...
+            bin.display(),
+            log.display(),
+        );
+        std::fs::write(&wrapper, wrapper_body)
+            .with_context(|| format!("failed to write {}", wrapper.display()))?;
+
+        let (schedule, modifier) = schtasks_cadence(every);
+        let output = Command::new("schtasks")
+            .args([
+                "/Create",
```
tenequm: **[significant]** `schtasks /Create` without an XML import inherits Task Scheduler's default `<Settings>`, which fight this job's purpose:
- `DisallowStartIfOnBatteries=true`: an unplugged laptop never starts the sync at all.
- `StopIfGoingOnBatteries=true`: an in-flight sync is killed when the user unplugs.
- `StartWhenAvailable=false`: no catch-up after wake. The `MINUTE`/`HOURLY` repetitions resume so the gap is bounded, but a `1d` schedule on a machine asleep at trigger time misses the day entirely - launchd fires once on wake, and the systemd backend sets `Persistent=true` for exactly this.
- The task runs interactively with a visible window: a console pops onto the desktop and taskbar every tick, for the whole sync. Probably the most user-visible issue in the backend.

Fix likely means creating via `/XML` with a hand-written settings block (`AllowStartIfOnBatteries`, `DontStopIfGoingOnBatteries`, `StartWhenAvailable`, `Hidden` or a windowless launch shim), rather than `/SC`/`/MO` flags.

packages/pond/src/schedule.rs
```diff
@@ -914,3 +917,317 @@ mod tests {
...
+            .output()
+            .context("failed to run schtasks /Create")?;
+        if !output.status.success() {
+            bail!(
+                "schtasks /Create failed: {}",
+                decode_console(&output.stderr).trim()
+            );
+        }
+        line(&format!(
+            "schedule active (task-scheduler, every {})",
```
tenequm: **[significant]** Three parity gaps with the unix backends on this path:
1. No already-scheduled short-circuit: `ScheduleCmd::Start`'s own doc promises "re-running with the same one is a no-op", and all three unix backends honor it by comparing the existing artifact body plus registration state before rewriting. This always rewrites the wrapper and re-runs `/Create /F`.
2. Hand-rolled output instead of `line(&render_state(&Active { .. }))` - the `paint("schedule", dim())` prefix and column alignment with `pond status` are lost.
3. No `logs   <path>  (pond schedule logs)` pointer after start, which every unix start path prints.

packages/pond/src/schedule.rs
```diff
@@ -914,3 +917,317 @@ mod tests {
         assert_eq!(ScheduleEvery::from_secs(123), None);
     }
 }
+
+#[cfg(windows)]
+mod windows {
```
tenequm, line 931: **[significant]** Roughly a third of this module is a verbatim copy of `mod unix`: `run`, `status_line`, `status_snapshot`, `enum State`, `render_state`, `logs`, and `log_path` are identical modulo dropped doc comments (`run` is byte-for-byte the same match block). The file already has the right precedent: `ScheduleSnapshot` lives in the parent module because two backends need it, and this PR itself widened `ScheduleEvery::from_secs` to `cfg(any(unix, windows))` rather than copying it. With a second backend the shared pieces now have their second caller - please hoist them to the parent module so the platform modules keep only what actually differs (registration, probe, cadence mapping).

packages/pond/src/schedule.rs
```diff
@@ -914,3 +917,317 @@ mod tests {
...
+        }
+        line(&format!(
+            "schedule active (task-scheduler, every {})",
+            every.label()
+        ))?;
+        Ok(())
+    }
+
+    fn stop() -> Result<()> {
+        let existed = task_exists();
```
tenequm: **[minor]** TOCTOU: if the task disappears between `task_exists()` and `/Delete` (a second `pond schedule stop`, or deletion via the Task Scheduler UI), this bails with `schtasks /Delete failed` instead of the intended "nothing was scheduled". Note the naive fix is wrong too - `/Delete /F` on a missing task exits non-zero rather than no-opping - so either keep the pre-check and treat a delete failure after a positive check as "already gone", or drop the pre-check and match the not-found error.

packages/pond/src/schedule.rs
```diff
@@ -914,3 +917,317 @@ mod tests {
...
+            .args(["/Query", "/TN", TASK_NAME])
+            .output()
+            .map(|output| output.status.success())
+            .unwrap_or(false)
+    }
+
+    /// The registered task's cadence, recovered from its XML so `pond status`
+    /// can name it. `None` when the task lacks a recognized interval (it is
+    /// still reported active).
+    fn read_interval() -> Option<ScheduleEvery> {
```
tenequm: **[minor]** Two things here: (1) `probe()` spawns schtasks twice - `task_exists` then this query. The unix module's design note ("one probe serving every `pond status` need") applies: a single `/Query /TN <name> /XML ONE` answers both existence (exit code) and cadence. (2) In the `if let / else if` chain below, a matched-but-unparsable first branch suppresses the later ones (the `?` applies to the combined result), and `between` anchors on the first `<Interval>PT` but scans the whole remainder for the close tag, so open and close can come from different elements. All five cadences pond registers resolve correctly today; it breaks the moment the XML carries a second `<Interval>`. Cheap to make robust while you are adding tests (see the comment on `from_secs`).

packages/pond/src/schedule.rs
```diff
@@ -50,7 +50,7 @@ impl ScheduleEvery {
         }
     }
 
-    #[cfg(unix)]
+    #[cfg(any(unix, windows))]
```
tenequm: **[significant]** `from_secs` was widened to `cfg(any(unix, windows))`, but the tests module below is still `#[cfg(all(test, unix))]` - so even this shared helper's test does not run on Windows, and the new backend ships zero tests overall. `between`, `read_interval`'s XML parsing, `schtasks_cadence`, and `decode_console` are all pure and trivially testable; the unix module tests exactly this class in exactly the right place (round-tripping `cron_entry` through `cron_entry_interval` for all five cadences). A `schtasks_cadence -> sample task XML -> read_interval` round-trip is the direct analog.

packages/pond/src/config.rs
```diff
@@ -832,12 +832,26 @@ pub fn contract_home_under(path: &Path, home: &Path) -> PathBuf {
...
 
-/// [`contract_home_under`] against the process `HOME`. Returns the input
+/// The user's home directory, resolved per-OS from the process environment.
+/// Windows reads `USERPROFILE` (the native home; `HOME`, when present at all,
+/// is a POSIX-style path set by git-bash/msys/Cygwin and points somewhere a
+/// native binary must not follow). Unix reads `HOME`. Empty values are treated
+/// as unset. Every home-relative default path and the adapter auto-discovery
+/// probes resolve through this, so pond finds `~/.claude`, `~/.omp`, and
+/// friends on Windows the same way it does on Unix.
+pub fn home_dir() -> Option<PathBuf> {
```
tenequm, line 872: **[significant]** Three things on `home_dir()`:
1. The doc wording that reached `adapter/mod.rs` ("`HOME`, then `USERPROFILE` on Windows") reads as a fallback chain, but the implementation is an exclusive per-OS switch - and this comment right here explains why the fallback reading would be the git-bash bug the function exists to avoid. Please align the adapter-side wording.
2. Test-harness gap: both subprocess sandboxes (`tests/integration/resume.rs:111`, `tests/integration/cli_stdout_contract.rs:25`) isolate the spawned pond with `.env("HOME", ...)` plus the XDG vars - which this function never reads on Windows. Everything not covered by an XDG var (adapter auto-discovery, `~` expansion, `contract_home`) then reads the developer's real `%USERPROFILE%`. Latent today, but please set `USERPROFILE` alongside `HOME` in those harnesses.
3. Optional: `std::env::home_dir()` is fixed and un-deprecated on our pinned toolchain, ignores `HOME` on Windows (your exact git-bash concern), and additionally falls back to the Win32 profile API when `USERPROFILE` is unset - this function could just delegate to it.

packages/pond/src/syncstate.rs
```diff
@@ -74,15 +78,14 @@ fn try_acquire_sync_lock_in(dir: &Path, store_key: &str) -> Result<SyncLockState
...
                 started_at: Utc::now(),
             };
-            // Best-effort holder info; the lock itself is the flock, not the bytes.
-            let _ = file.set_len(0);
-            let _ = file.seek(SeekFrom::Start(0));
-            let _ = serde_json::to_vec(&holder).map(|bytes| file.write_all(&bytes));
-            let _ = file.flush();
+            // Best-effort breadcrumb; the lock itself is the flock, not the bytes.
+            if let Ok(bytes) = serde_json::to_vec(&holder) {
+                let _ = std::fs::write(&holder_path, bytes);
```
tenequm: **[minor]** The sibling-file move itself is sound (and not a regression - the old code never cleared holder bytes on release either). Three small things: (1) the `SyncLockHolder` doc above still says the holder is "written into the lock file on acquire"; (2) this is a plain `std::fs::write` where `write_last_sync_in` in this same file deliberately uses temp+rename "so a concurrent `pond status` never reads a half-write" - same file, same concern, same fix; (3) one `.holder.json` per store key now accumulates in the state dir forever - a best-effort `remove_file` in the guard's drop would keep it tidy.

docs/spec.md
```diff
@@ -109,7 +109,10 @@ These are stable positions. pond will not:
 
 ### 2.4 Platform
 
-Linux and macOS. Windows is not in v1 scope.
+Linux, macOS, and Windows. On Windows the vendored protoc build (`protobuf-src`)
```
tenequm, line 112: **[minor]** Two spec surfaces this rewrite should also cover:
1. This caveat list omits the most consequential Windows caveat the spec already encodes: `local-store-durability` (section 3.3) is unix-only and "Windows relies on `local-store-self-heal`" - for a data store that belongs in the platform section.
2. Section 7.8's `pond schedule` entry still names only launchd/systemd/cron and carries the `XDG_STATE_HOME` pinning sentence - it needs the Task Scheduler backend added (and the pinning honored, per the `schedule.rs` comment). The `schedule.rs` module header has the same launchd/systemd/cron-only enumeration.

README.md
```diff
@@ -86,6 +86,8 @@ cargo install --path packages/pond --features cuda
 
 On macOS the Metal backend is selected automatically; on other systems the CPU fallback runs without extra features.
 
+On Windows, the Homebrew/Nix packages do not apply; install with `cargo install pond-db` or build from source. Both need a protobuf compiler (`protoc`) on `PATH` first (`winget install Google.Protobuf`), because the vendored protoc build used on Unix does not link with MSVC. Embeddings run on the CPU backend.
```
tenequm: **[minor]** Two more stale surfaces once Windows is claimed at line 63: the "Scheduled sync" section below still says "launchd on macOS, systemd user timers or cron on Linux", and `docs/site/src/pages/get-started/install.mdx` still opens with "macOS and Linux. Windows is out of v1 scope" (frontmatter description too). Also note this paragraph's protoc rationale changes if the Cargo.toml gate becomes MSVC-only as suggested - `cargo install` with the MSVC toolchain still needs a system protoc, but the binary we publish is windows-gnu and unaffected.

packages/pond/src/main.rs
```diff
@@ -1097,8 +1097,36 @@ fn try_raise_fd_limit(_target: u64) -> anyhow::Result<()> {
...
+/// platform.
+fn main() -> anyhow::Result<()> {
+    const MAIN_STACK_SIZE: usize = 16 * 1024 * 1024;
+    let worker = std::thread::Builder::new()
+        .name("pond-main".into())
+        .stack_size(MAIN_STACK_SIZE)
+        .spawn(|| {
+            let runtime = tokio::runtime::Builder::new_multi_thread()
+                .enable_all()
+                .thread_stack_size(MAIN_STACK_SIZE)
```
tenequm, line 1116: **[nit]** Non-blocking: `thread_stack_size` also applies to tokio's blocking pool (512 threads max), so worst case is ~8 GiB of reserved address space. Reserved, not committed, so this is fine on x86_64 - but since `block_on` polls the CLI future on the dedicated 16 MiB thread itself, the runtime-wide override may not be needed at all. If it stays, `.max_blocking_threads(64)` would be cheap polish. The dedicated-thread approach itself is right, and behaviorally equivalent on unix as far as I could verify (exit codes, signals, panic path).

tenequm (changes requested):
Thanks for this - it is a careful, well-argued PR. The per-OS branching genuinely stays shared-path rather than becoming a Windows fork, the WHY comments are substantive, and raising the spec section 2.4 scope question openly instead of quietly rewriting it is appreciated.

I verified the suite on macOS (fmt, clippy `-D warnings`, and all tests green; `Cargo.lock` genuinely unchanged as claimed) and went through the Task Scheduler backend line by line. The inline comments carry the specifics; the short version is three blockers, everything else is smaller:

1. **The `cfg(unix)` protoc gate breaks our existing Windows release build.** The release pipeline already cross-compiles `x86_64-pc-windows-gnu` with cargo-zigbuild on a Linux runner (`moon.yml` `build-dist`) and publishes `pond-x86_64-pc-windows-gnu.exe` - today that build works precisely because the `protoc` feature vendors protoc. Details and the fix (`cfg(not(target_env = "msvc"))`) inline on `Cargo.toml`. Also worth knowing for your testing: the binary we ship is windows-gnu, not MSVC - which likely means your dedicated-thread stack fix repairs the already-published `pond.exe` too, a nice bonus.
2. **The Task Scheduler registration does not pin `XDG_STATE_HOME`**, which `docs/spec.md` section 7.8 makes a contract - a shell-only override splits the single-flight lock between scheduled and manual syncs. Inline on `schedule.rs`.
3. **`cmd /C` receives MSVCRT-escaped arguments, which cmd does not parse by those rules** - a `_command` containing quotes gets mangled, and quoting is mandatory the moment a vault or item name has a space. Inline on `substrate.rs`.

On the CI job: thanks for dropping it per my earlier comment - please also update the PR body and the commit message, which still describe a `windows-latest` job that is no longer in the diff.

On the section 2.4 scope question and your offer to split into stages: I will give you a definitive answer after a fuller pass tomorrow; the findings here are the correctness direction so you are not blocked in the meantime. One design question I would like your input on while I think about scope: the current `home_dir()` approach means the store lands at `C:\Users\x\.local\share\pond` and config at `~\.config\pond\config.toml` - unix dotdirs under the profile rather than `%LOCALAPPDATA%` / `%APPDATA%`. Both are defensible, but pre-1.0 is the only cheap moment to pick, because changing later strands existing Windows stores. If you have a view, I would like to hear it.

Things I checked and cleared, so you know they were examined rather than skipped: the `main()` dedicated-thread change is behaviorally equivalent on unix (exit codes, signal handling, panic path all match); the `.holder.json` move is not a regression (the old code never cleared holder bytes on release either, and the breadcrumb is only read while a live holder exists); `drop(store)` in the rowmap test removes nothing the assertions relied on; the strip-one-newline path already handled CRLF, so the `cmd /C` output handling is correct; the schtasks exit-code probe is locale-safe and the UTF-16 console decoding is a nice touch.

One process ask: **please no more force-pushes from here on** - push follow-up commits instead. The branch was force-pushed while review was in progress, which invalidates in-flight review state and makes it impossible to see what changed between versions. We squash-merge anyway, so extra commits on the branch are free.

packages/pond/Cargo.toml
BadAd84: Done - changed to `cfg(not(target_env = msvc))` in both `[dependencies]` and `[dev-dependencies]`. The redundant `cfg(unix)` dev-dep `lance` re-declaration is replaced by the same non-MSVC gate.

packages/pond/src/schedule.rs
BadAd84: Done. The wrapper now opens with `set XDG_STATE_HOME=<state_root>` where the value is `state_root()` (not `pond_state_dir()`, which would double the path). `%` in embedded paths is escaped as `%%` for the batch context; a hard bail fires if `state_root()` contains `%` - Task Scheduler expands `%VAR%` in `<Arguments>` at runtime with no escape syntax, so the path needs to be clean at registration (mirrors the unix gate that rejects `%` for the same reason in cron/plist templates). Wrapper is written after `/Create` succeeds.

packages/pond/src/schedule.rs
BadAd84: Done via `/Create /XML` with a `<Settings>` block: `DisallowStartIfOnBatteries=false`, `StopIfGoingOnBatteries=false`, `StartWhenAvailable=true`, `Hidden=true`.

On the console window: `<Hidden>true</Hidden>` only hides the task in the Task Scheduler UI - it does not suppress the console. Addressed with a `.vbs` shim alongside the `.cmd` wrapper: `WScript.Shell.Run(..., 0, False)` with WindowStyle=0. The task Action is `wscript.exe //B //NoLogo <shim.vbs>`; the shim calls the `.cmd` invisibly.

One additional fix found during verification: the XML temp file must be written as UTF-16LE with BOM (the encoding the `encoding=UTF-16` declaration names and Task Scheduler's own export format). Writing it as plain UTF-8 caused non-ASCII state paths to register via the ANSI codepage as mojibake - the task shows registered but silently does nothing on every tick.

packages/pond/src/schedule.rs
BadAd84: Done: (1) already-scheduled no-op checks both artifact body hashes plus the probed cadence, matching the launchd/systemd/cron pattern. (2) output via `render_state(&Active { .. })`. (3) logs path printed after registration.

packages/pond/src/schedule.rs
BadAd84: Done. `State`, `render_state`, `log_path`, `status_line`, `status_snapshot`, `run`, and `logs` are hoisted to the parent module. `systemd_timer_enabled` is `pub(super)` so the shared `logs()` can delegate to journalctl on Linux+systemd. Platform modules keep only `start`, `probe`, and `stop`.

packages/pond/src/schedule.rs
BadAd84: Done. `/Delete /F` runs unconditionally; if it fails a subsequent `probe()` call disambiguates: `Inactive` means already gone (race or never registered), `Active` means a genuine failure. Locale-safe - no stderr text parsing.

packages/pond/src/schedule.rs
BadAd84: Done: (1) single `/Query /TN <name> /XML ONE` call covers both existence (exit code) and cadence (XML body). (2) `parse_interval_from_xml` uses `strip_prefix`/`strip_suffix` instead of the `between()` chain, so a matched-but-unparsable first branch no longer suppresses the others.

packages/pond/src/schedule.rs
BadAd84: Done. `every_round_trips_through_secs_and_labels` now runs on all platforms (was `cfg(all(test, unix))`). Windows module tests added: `all_cadences_round_trip_through_task_xml` (all five cadences through `task_xml` -> `parse_interval_from_xml`), `between_finds_content_between_delimiters`, `decode_console_handles_utf8_and_utf16le_bom`, `task_xml_uses_wscript_and_contains_expected_settings`, `vbs_body_has_windowstyle_zero_and_quotes_wrapper`.

packages/pond/src/syncstate.rs
BadAd84: Done: (1) doc corrected - holder is in the sibling `.holder.json` file, not the lock file itself. (2) holder write uses temp+rename. (3) `SyncLockGuard::drop` removes the `.holder.json` best-effort.

docs/spec.md
BadAd84: Done: (1) section 2.4 notes `local-store-durability` is unix-only and Windows relies on `local-store-self-heal` (directory-fsync semantics differ). (2) section 7.8 adds Task Scheduler alongside launchd/systemd/cron, including the `XDG_STATE_HOME` pinning note. Module header updated.

packages/pond/src/main.rs
BadAd84: Added `.max_blocking_threads(64)`.

packages/pond/src/substrate.rs
BadAd84: Fixed with `CommandExt::raw_arg` on the cmd branch, passing the command string verbatim as `/C <command>`. Shell resolves from `COMSPEC` with `cmd` fallback. The existing test was extended with a case carrying both a space and a quote (`echo "hello world"`), with cfg-gated expected output (cmd preserves the outer quotes; sh strips them).

packages/pond/src/config.rs
BadAd84: Done: (1) `adapter/mod.rs` doc now says "USERPROFILE on Windows, HOME on Unix" rather than implying a fallback chain. (2) Both integration harnesses now pin `APPDATA` and `LOCALAPPDATA` alongside `HOME`/`USERPROFILE`/`XDG_*` so the spawned process is fully sandboxed. On `std::env::home_dir()`: left as the explicit switch.

On the storage location question: went Windows-native - `%LOCALAPPDATA%\pond` for data and cache, `%APPDATA%\pond\config.toml` for config.

README.md
BadAd84: Done: "Scheduled sync" section mentions Task Scheduler. `install.mdx` removes "Windows is out of v1 scope" and adds a `cargo binstall pond-db` block (prebuilt windows-gnu binary, no `protoc` needed). README Windows install updated to recommend binstall and correct the protoc note - only `cargo install` on an MSVC toolchain needs a system `protoc`; the published binary is windows-gnu and vendors it automatically.

packages/pond/src/config.rs
```diff
@@ -474,28 +474,51 @@ impl Default for EmbeddingsConfig {
...
+/// On other platforms: `$XDG_DATA_HOME/pond`, then `$HOME/.local/share/pond`,
+/// then `.pond` as a last resort. `xdg_data_home` is honored only if
+/// absolute, per the XDG base-directory spec.
 pub fn default_storage_path(xdg_data_home: Option<PathBuf>, home: Option<PathBuf>) -> Result<Url> {
+    #[cfg(windows)]
+    if let Some(local_app_data) = std::env::var_os("LOCALAPPDATA")
+        .map(PathBuf::from)
+        .filter(|p| p.is_absolute())
+    {
+        return url_for_path(local_app_data.join("pond"));
```
tenequm: **Directory layout (1/2).** This makes `%LOCALAPPDATA%\pond` the store root while `default_cache_path` below puts the cache at `%LOCALAPPDATA%\pond\cache` - a foreign subdirectory *inside* the Lance store root. Please split one root three ways: data -> `%LOCALAPPDATA%\pond\data`, cache -> `%LOCALAPPDATA%\pond\cache`, state -> `%LOCALAPPDATA%\pond\state` (see the `syncstate.rs` comment); config stays `%APPDATA%\pond\config.toml` as you have it. That keeps one `pond` root per user with the same three-role separation the unix layout has. And deliberately not Roaming for data/state: a roaming profile syncs `%APPDATA%` at logon, and a Lance store must never ride that.

packages/pond/src/config.rs
```diff
@@ -474,28 +474,51 @@ impl Default for EmbeddingsConfig {
...
+    }
+    #[cfg(not(windows))]
```
tenequm: **XDG override consistency.** This gate (and its twins in `default_cache_path` / `default_config_path`) hard-ignores an explicitly-set `XDG_*` var on Windows - while `state_root()` in `syncstate.rs` still honors `XDG_STATE_HOME` first, so the override behavior is split across the four dirs. Please make them consistent: honor an absolute `XDG_*` override when set on Windows too, with the native dirs as the fallback leg. That also keeps the sandboxed test harnesses meaningful cross-platform (they pin `XDG_*` alongside `APPDATA`/`LOCALAPPDATA`).

packages/pond/src/syncstate.rs
```diff
@@ -20,7 +19,7 @@ pub(crate) fn state_root() -> PathBuf {
     std::env::var_os("XDG_STATE_HOME")
         .map(PathBuf::from)
         .filter(|path| path.is_absolute())
-        .or_else(|| std::env::var_os("HOME").map(|home| PathBuf::from(home).join(".local/state")))
+        .or_else(|| crate::config::home_dir().map(|home| home.join(".local/state")))
```
tenequm: **Directory layout (2/2).** With `home_dir()` this now resolves state to `%USERPROFILE%\.local\state\pond` on Windows - a unix dotdir in the profile root. The native home for it is `%LOCALAPPDATA%\pond\state` (`dirs::state_dir()` is `None` on Windows by design; LOCALAPPDATA is the state analog). `XDG_STATE_HOME` should keep winning when set, exactly as this line already does.

tenequm (changes requested):
Thanks for the thorough revision round - this is solid work. Full transparency on process: after your PR came in, I wanted an unbiased second opinion before reviewing the design decisions, so I worked out a Windows-support plan from scratch deliberately without reading your diff, and only then compared the two. That plan is now in the repo: [docs/plans/2608-13-windows-support-plan.md](https://github.com/tenequm/pond/blob/main/docs/plans/2608-13-windows-support-plan.md). It and your PR agree on essentially the whole runtime core (the protoc gating, per-OS home resolution, the native dirs, `raw_arg` on the cmd path, the lock-holder sidecar) - strong mutual validation. One find deserves particular credit: the main-thread stack overflow. The clean-room plan missed it entirely; your dedicated-thread fix is exactly right, and it's now folded into the doc, credited to this PR.

Direction confirmed, and I'd like to land this as a single PR - no need for the staged split you kindly offered. Two asks remain for this final round, both about durable on-disk state that's cheap to fix now and costly once users accumulate data - see the line comments on `config.rs` and `syncstate.rs`.

Everything else in the current revision is accepted as-is. After the merge, the remaining Windows work continues per the plan doc above (so none of it lands on you, and none of it should read as scope creep later):

- I'm walking back my earlier "no need for a CI job" comment - that was wrong. The published windows-gnu binary has never been executed by any CI, so I'll add a native `windows-2025` job that runs the full suite and builds/ships an msvc artifact, dropping the gnu one. At that point the `cfg(not(target_env = "msvc"))` gate simplifies to `cfg(not(windows))` - no action needed from you.
- The `.vbs` shim works today, but VBScript is deprecated and on Microsoft's removal path, so I'll later swap the scheduled action to a small windowless `pondw.exe` (`windows_subsystem = "windows"`). Your task XML, settings block, and probe logic all carry over unchanged.
- The remaining gaps (PATHEXT-aware `claude`/`codex` detection and `.cmd` shim spawning in `pond init`, a cliclack bump for Ctrl-C inside prompts, adapter path decoding for claude-code/claude-desktop) are queued in the doc's follow-up series.

Thanks again - the 331-session end-to-end verification is exactly the kind of evidence that makes a port like this reviewable.

packages/pond/src/config.rs
BadAd84: Done. `default_storage_path` now returns `%LOCALAPPDATA%\pond\data` (not the bare `\pond` root). The layout is:

```
%LOCALAPPDATA%\pond\data    - Lance datasets
%LOCALAPPDATA%\pond\cache   - rowmap cache
%LOCALAPPDATA%\pond\state   - lock, last-sync, scheduler log
%APPDATA%\pond\config.toml  - config (Roaming; data/state stay Local)
```

packages/pond/src/config.rs
BadAd84: Done. All three functions now check `XDG_DATA_HOME`/`XDG_CACHE_HOME`/`XDG_CONFIG_HOME` first (honoring an absolute override on every platform), with the native Windows dirs as the fallback. This matches `state_root()`, which already honored `XDG_STATE_HOME` first. The test harnesses pin both `XDG_*` and `APPDATA`/`LOCALAPPDATA`, so they stay hermetic regardless of which path the function takes.

packages/pond/src/syncstate.rs
BadAd84: Done. `state_root()` native Windows fallback is now `%LOCALAPPDATA%\pond\state`. `pond_state_dir()` on Windows returns `state_root()` directly with no `\pond` suffix - whether `XDG_STATE_HOME` is set by the user or pinned by the scheduler wrapper. The scheduler sets `XDG_STATE_HOME=state_root()` in the `.cmd`, so the scheduled job resolves the same dir as a manual run. On Unix the `\pond` suffix is retained (one app under the shared XDG state home).

tenequm (approved):
Merged. Thanks for this - it was a genuinely well-built PR, and the revision rounds were fast and precise.

A few things that stood out, since they deserve naming:

- **The main-thread stack overflow.** I had worked out a Windows plan from scratch before reading your diff, specifically to get an unbiased second opinion, and it missed this entirely. A binary that can't even run `--version` on MSVC is the kind of thing you only find by actually running it. Your dedicated-thread fix is the right shape, and it's now in the plan doc credited to this PR.
- **The UTF-16LE BOM discovery.** `schtasks /Create /XML` reading a BOM-less file through the ANSI code page, registering successfully, and then silently doing nothing on every tick is a nasty failure mode, and it is undocumented by Microsoft. I went and verified the encoding matrix independently and you were right on every point.
- **The 331-session end-to-end verification.** That is what made a port this size reviewable at all.

Fourteen findings addressed in one round, then the directory-layout and XDG-override asks in a second, with the `pond_state_dir` suffix consequence spotted and handled on your own initiative. That's a lot of careful work.

What happened since, so none of it comes as a surprise - I took the rest of the Windows work from here:

- A native `windows-2025` CI job now runs the full suite on every pull request and builds the shipped MSVC artifact, replacing the cross-compiled windows-gnu binary that no CI had ever executed. That's my earlier "no need for a CI job" comment walked back properly, and the `cfg(not(target_env = "msvc"))` gate is now simply `cfg(not(windows))`.
- The remaining gaps - PATHEXT-aware client detection, a network-path gate, adapter path decoding - have landed as well, and all four adapters plus your Task Scheduler backend are now verified against real captures on a Windows 11 machine.
- The `.vbs` shim still works and still ships; VBScript is on Microsoft's removal path, so it will later become a small windowless `pondw.exe`. Your task XML, settings block, and probe logic carry over unchanged.

You're credited in the changelog for the next release.

PR #154, opened 2026-08-14
fix(sync): release the sync lock explicitly on drop

### LINKED ISSUE #153: Rare flake: sync_lock_excludes_a_sibling_and_frees_on_drop fails under the full bin suite
`syncstate::tests::sync_lock_excludes_a_sibling_and_frees_on_drop` intermittently fails on the final assertion - the re-acquire after `drop(guard)` reports `Busy` instead of `Acquired`:

```
thread 'syncstate::tests::sync_lock_excludes_a_sibling_and_frees_on_drop' panicked at packages/pond/src/syncstate.rs:239:9:
assertion failed: matches!(try_acquire_sync_lock_in(dir.path(), "k1").unwrap(),
    SyncLockState::Acquired(_))
test result: FAILED. 37 passed; 1 failed
```

## Reproduction rate

Observed once during a full `cargo test --locked` on `aarch64-apple-darwin` (macOS 15, rustc 1.95.0). It does not reproduce in isolation or on repeat:

- `cargo test --bin pond syncstate::tests::sync_lock_excludes` x10: 10/10 pass
- full `cargo test --bin pond` x15: 15/15 pass

So the trigger involves the full bin suite's parallel load, not the test's own logic, and the rate is roughly 1 in 16 or rarer.

## Why it matters now

The `windows-verify` gate added in #152 runs the full suite on every release PR and is a required check. A ~1-in-16 flake in a required gate blocks releases at a rate that will be indistinguishable from a real regression, and Windows makes the underlying mechanism worse rather than better: `File::try_lock` maps to `LockFileEx` there, whose byte-range locks are mandatory rather than advisory like the `flock` used on Apple targets.

## What is ruled out

The lock file lives in a per-test `TempDir`, so this is not contention with a real `pond sync` on the developer's machine, and not cross-test path collision. On Apple targets `std::fs::File::try_lock` uses `flock(2)` (`library/std/src/sys/fs/unix.rs`), whose locks belong to the open file description - the second `File::open` in the `Busy` branch closing its own descriptor should not release the first one's lock. The mechanism is not yet explained.

## Suggested next step

Reproduce under `--test-threads` pressure with the lock syscalls traced before changing anything. A blind retry loop in the test would hide whichever of the two possibilities this is: a test-harness artifact, or a real gap in the guard's release semantics that also affects concurrent `pond sync` invocations.

tenequm:
Fixes #153, which I filed while building the Windows CI gate (#152) after a rare failure of `sync_lock_excludes_a_sibling_and_frees_on_drop`. It turned out not to be a flaky test - it is a real bug the test was correctly catching.

## The bug

`flock` belongs to the open file *description*, not to the descriptor. Any subprocess spawned while `SyncLockGuard` is alive inherits a duplicate descriptor, so dropping our `File` did **not** release the lock - it stayed held until that child execed (CLOEXEC) or exited.

A later acquire then reports `Busy`, and with `holder: None`, because `Drop` had already removed the holder sidecar. So the symptom is a sync that refuses to start - or, under `--no-wait`, silently skips - blamed on a holder it cannot name, because of an unrelated child process.

## Why it is not just a test artifact

`pond serve` holds this lock across an in-serve sync (`main.rs:3886`) while continuing to serve requests, and credential `_command` sources spawn `sh -c` (`substrate.rs:619-625`). A spawn racing the guard's drop is exactly the shape above. The test suite hit it because it spawns subprocesses throughout; the fork window is microseconds, which is why the rate was roughly 1 in 16 runs.

## The fix

`Drop` now calls `File::unlock()` before closing. `LOCK_UN` clears the description's lock however many descriptors still reference it, so release is deterministic instead of dependent on every inherited copy being closed. It stays best-effort - an unlock failure just leaves the previous close-time behavior.

## How it was found and verified

Reasoning did not produce the mechanism, so I went empirical:

- 110,000 plain acquire/drop/reacquire cycles (30k single-threaded, 8x10k concurrent): **zero reproductions**. The lock cycle itself was never the problem.
- The same cycle with a thread spawning subprocesses alongside: **reproduced at round 5.**

The regression test is that second shape. It is a true regression test, not a vacuous one - with the `unlock()` line removed it fails at round 3; with it, 20,000 rounds pass. It is `cfg(all(test, unix))`: the mechanism is `flock` plus fd inheritance across fork, and Windows mandatory byte-range locks have different handle-inheritance semantics, so the same shape would prove nothing there.

## Why this mattered enough to fix now

`windows-verify` (#152) runs the full suite on every release PR as a required check. A ~1-in-16 failure in a required gate is indistinguishable from a real regression and would have blocked releases at that rate.

Full suite green locally: 293 lib + 39 bin + 87 integration.

### DIFF
--- a/packages/pond/src/syncstate.rs
+++ b/packages/pond/src/syncstate.rs
@@ -77,17 +77,28 @@ pub(crate) enum SyncLockState {
     Busy(Option<SyncLockHolder>),
 }
 
-/// Held for the whole sync run. The OS drops the flock when the file closes,
-/// so a killed sync can never leave a stale lock. The lock file is never
-/// unlinked - unlink while a sibling holds the path open hands out two locks.
-/// `SyncLockGuard::drop` removes the sibling `.holder.json` best-effort.
+/// Held for the whole sync run. `Drop` unlocks explicitly, and the OS drops the
+/// flock again when the file closes, so a killed sync can never leave a stale
+/// lock. The lock file is never unlinked - unlink while a sibling holds the path
+/// open hands out two locks. `SyncLockGuard::drop` also removes the sibling
+/// `.holder.json` best-effort.
 pub(crate) struct SyncLockGuard {
-    _file: File,
+    file: File,
     holder_path: PathBuf,
 }
 
 impl Drop for SyncLockGuard {
     fn drop(&mut self) {
+        // Release explicitly instead of relying on close. `flock` belongs to the
+        // open file *description*, not the descriptor: any subprocess spawned
+        // while this guard was alive inherited a duplicate, so dropping our
+        // `File` leaves the lock held until that child execs or exits. The
+        // window is small but real - it makes a later acquire report `Busy` with
+        // no holder, so a sync refuses to start (or, under `--no-wait`, silently
+        // skips) because of an unrelated child process. `unlock` clears the
+        // description's lock outright, however many descriptors still reference
+        // it. Best-effort: an unlock failure leaves the old close-time behavior.
+        let _ = self.file.unlock();
         // Best-effort: the file may already be absent (concurrent stop, etc.).
         let _ = std::fs::remove_file(&self.holder_path);
     }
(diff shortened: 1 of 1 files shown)


tenequm:
Folded into #152 - the lock fix belongs with the Windows gate that surfaced it, not in a separate PR.

PR #157, opened 2026-08-14
ci(windows): put the kache store on D: and uncap prefetch

tenequm:
`windows-verify` swings between **17m54s and 51m02s** on an unchanged dependency tree. This changes two things and measures the result.

## What the job's own kache report shows

Across five runs on `feat/windows-adapter-paths`, the same expensive crates flip between hit and miss with no change to `Cargo.lock`:

| Run | Test step | Hit rate | Compiled | Top misses |
|---|---|---|---|---|
| `e780a50` | 22m38 | 98.7% | 10 | `lance`, `lance_index`, `lance_namespace_impls` |
| `43aa072` | 17m54 | 97.5% | 20 | `datafusion_optimizer`, `datafusion_sql`, `datafusion` |
| `26d27e5` | **51m02** | 91.6% | 67 | `lance` **and** the datafusion set **and** `candle_transformers` |
| `2163564` | 38m45 | 90.7% | 74 | same shape |

`lance` missed, then hit, then missed again. `GC: 0 entries evicted` rules out local eviction, and the commits in between touch only `packages/pond/**`, so it is not key instability.

## Cause 1: prefetch is capped far below this graph

kache defaults `KACHE_PREFETCH_MAX_BYTES` to **2GiB** and `KACHE_PREFETCH_DEADLINE_SECS` to **300**. This build is ~790 crates where `lance` alone is 287 MB. Prefetch stops well short, and what it drops is whatever it did not reach - biasing hard toward the largest artifacts. That matches the reported `Prefetch: 0/719 hits (0.0%)` and the size-correlated miss list. Both are set to `0` (unlimited) here.

## Cause 2: the store is on the slow, small disk

kache defaults to `%LOCALAPPDATA%\kache` on `C:`. Per [actions/runner-images#12416](https://github.com/actions/runner-images/issues/12416), `C:` has ~29 GB free and is network-attached, while `D:` has ~147 GB of fast local storage and already holds `target/` (`D:\a\pond\pond`). `KACHE_CACHE_DIR` moves the store to `D:`, which also removes the cross-volume copy the job warns about on every run, and is the precondition for a future `KACHE_WINDOWS_HARDLINK` restore.

## Baseline to beat

```
Test step:         38m 45s
Hit rate:          90.7% count / 76.7% by compile cost (719/793, 74 compiled)
Miss compile work: ~46 min aggregate
Cache hit overhead: ~7 min aggregate, avg 590ms/hit
Prefetch:          0/719 hits (0.0%)
```

Scoped to `windows-verify` only - `windows-dist` is release-gated and keeps the current settings until this is proven.

## Not done here

- `sync: true` on kache-action, which pulls up front via `kache sync --pull`. Kept in reserve: it is the blunt version of the same idea, and worth trying only if uncapping prefetch does not land.
- `KACHE_WINDOWS_HARDLINK`, which needs same-volume (this PR) but carries a corruption risk if a build rewrites a restored output in place.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -165,11 +165,17 @@ jobs:
   # The shipped msvc zip. Release-gated rather than per-PR: it produces an
   # artifact nobody consumes outside a release, and it is the only thing in the
   # pipeline that ever executes a Windows binary.
+  # `!cancelled()` is load-bearing: release-prep only runs on a main push, and a
+  # skipped `needs` skips its dependents, so a bare `needs:` would stop this ever
+  # running on the release PR.
   windows-dist:
+    needs: [release-prep]
     if: >-
-      github.event_name == 'workflow_dispatch' ||
-      (github.event_name == 'push' && github.ref == 'refs/heads/main') ||
-      (github.event_name == 'pull_request' && startsWith(github.head_ref, 'release-plz-'))
+      !cancelled() && (
+        github.event_name == 'workflow_dispatch' ||
+        (github.event_name == 'pull_request' && startsWith(github.head_ref, 'release-plz-')) ||
+        needs.release-prep.outputs.is_release == 'true'
+      )
     runs-on: windows-2025
     timeout-minutes: 180
     # Not cancellable, same rule as dist-build: this produces a release artifact
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -127,6 +127,17 @@ jobs:
     # REPLACES the +crt-static rustflags from .cargo/config.toml.
     env:
       BAZEL_REMOTE_AUTH: ""
+      # kache defaults its store to %LOCALAPPDATA% on C:, which on this image is
+      # a slow ~29 GB network-attached disk; D: is ~147 GB of fast local storage
+      # and already holds target/ (actions/runner-images#12416). Same volume as
+      # the build tree is also the precondition for any future hardlink restore.
+      KACHE_CACHE_DIR: D:\kache
+      # Prefetch stops at 2GiB of blobs or 300s of scheduling by default, and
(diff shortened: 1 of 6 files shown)


tenequm:
## Measured: 38m45s -> 9m55s

First run on this branch, against the baseline recorded in the PR body:

| | Baseline (`2163564`) | This PR (`66a42c8`) |
|---|---|---|
| Test step | 38m 45s | **9m 55s** |
| Hit rate | 90.7% count / 76.7% by compile cost | **99.7% count / 98.8% by compile cost** |
| Crates compiled | 74 | **2** |
| Miss compile work | ~46 min | **~3 min** |
| Cache hit overhead | ~7 min, avg 590ms/hit | ~6 min, avg 430ms/hit |
| Compile work avoided | ~2.5h | ~3.6h |

**The uncapped prefetch is what collapsed the misses.** 74 compiled crates went to 2, and the by-compile-cost hit rate - the number that actually predicts wall clock - went from 76.7% to 98.8%. The 2GiB/300s caps were dropping precisely the expensive crates, exactly as the miss lists suggested.

**The `D:` move shows up as lower per-hit overhead** (590ms -> 430ms) and the disappearance of the cross-volume advisory from the log, which had been firing on every run.

### One thing that did not change

`Prefetch: 0/791 hits (0.0%)` - still exactly zero, on a run where the cache clearly served 791 of 793 crates. That confirms the metric is misattributed rather than measuring anything real, consistent with [kunobi-ninja/kache#618](https://github.com/kunobi-ninja/kache/issues/618) (prefetch telemetry cannot compute byte precision or blocking time) and the earlier [#581](https://github.com/kunobi-ninja/kache/issues/581) counter bug. Worth an upstream report with this run as evidence: the caps were real and load-bearing, while the counter that would have shown them working reads zero either way.

`sync: true` stays unnecessary - the blunt version of a problem that turned out to be two config caps.

PR #198, opened 2026-08-30
fix(serve): stop cleanly when a supervisor asks

### LINKED ISSUE #195: pond serve never stops when a supervisor asks it to
Follow-up to #194. Two separate problems with the same symptom: a supervised
`pond serve` is always killed rather than stopped. Both reproduce against the
v0.16.3 release binary in a container, `serve --transport http --host 0.0.0.0`,
with pond as PID 1.

## 1. SIGTERM is ignored

`transport.rs::shutdown_signal` awaits `tokio::signal::ctrl_c()`, which on unix
is SIGINT alone. Container runtimes, systemd and orchestrators all stop a
process with SIGTERM, and a process running as PID 1 gets no default
disposition for a signal it has not registered a handler for, so the signal is
discarded.

Measured, same binary, same container, only the signal differs:

```
SIGINT   exits in 116 ms, exit code 0
SIGTERM  nothing happens; SIGKILL at the end of the grace period, exit code 137
```

That is the whole stop timeout burned on every deploy, restart and scale event,
and under `--with-sync` the sync loop is killed outright instead of being
dropped cleanly.

The fix is small: also await
`tokio::signal::unix::signal(SignalKind::terminate())`, `select!` over the two,
`ctrl_c` alone on non-unix. I have it on a branch and will send it as a PR if
you want it.

## 2. Graceful shutdown never completes with an MCP client connected

Not signal-specific: this one bites on SIGINT too, which pond does handle.

`axum::serve(...).with_graceful_shutdown(...)` waits for every in-flight
connection to close, and an MCP streamable-HTTP client holds a `GET /mcp` SSE
stream open for the life of its session by design. The drain therefore never
finishes.

Measured, one client attached (`initialize`, `notifications/initialized`, then
the stream held open), SIGINT to pond:

```
t+5s   still running
...
t+90s  still running, no exit
```

The same build with no client attached exits in 116 ms. So on the long-lived
serve topology in #165, every restart ends in SIGKILL as soon as one agent is
connected. Together with 1, a hosted pond is never stopped cleanly at all.

Two ways out, and I did not want to pick one for you:

- Cancel rmcp's own `StreamableHttpServerConfig::cancellation_token` when the
  shutdown signal fires. It is documented as terminating active sessions, which
  should let the drain finish by itself. This is the transport's intended hook,
  but it needs the `CancellationToken` type, so either a direct `tokio-util`
  dependency (already in the tree via rmcp) or an rmcp re-export.
- Bound the drain: after the signal, give connections a few seconds and exit
  regardless. Blunter, no new dependency, and it also covers any future
  long-lived route.

Maziar110:
## Summary

Closes #195, both parts, in the shape you asked for.

**SIGTERM.** `shutdown_signal` awaited `ctrl_c` only, which on unix is SIGINT.
It now selects over SIGINT and SIGTERM, and falls back to ctrl-c alone (with a
warning) if the SIGTERM handler cannot be installed. `cfg(not(unix))` is
unchanged.

**The drain.** Option 1, as you picked. `serve` builds a `CancellationToken`,
hands it to `StreamableHttpServerConfig::with_cancellation_token`, and cancels
it when the stop signal fires, so live MCP sessions are torn down and axum's
graceful shutdown can finish on its own. A 5 second bounded drain sits behind
that as the backstop, so a client that ignores the close cannot hold the
process open either.

`tokio-util` is now a direct dependency. It was already in the tree via rmcp,
so the lockfile gains exactly one line and no version moves.

## Shape of the change

`serve` splits in two. It still binds, logs, and owns the signal handler, then
hands off to `serve_with_shutdown(listener, state, allowed_hosts, stop)`, which
takes the stop trigger as a future. That is what lets the test drive a real
socket and a real MCP stream without raising a process signal inside the test
binary.

## Test

`shutdown_completes_while_an_mcp_stream_is_open` in
`tests/integration/transport_http.rs`, over a real loopback socket rather than
`oneshot` against the `Router`, because the hang was in the connection drain
and `oneshot` never binds one.

It initializes an MCP session, opens `GET /mcp`, asserts the response head is
a 200 `text/event-stream`, then holds the socket unread and unclosed across the
shutdown and asserts `serve_with_shutdown` returns.

The 200 assertion is there because the test is easy to make vacuous: if the
stream were ever refused, the connection would close and shutdown would finish
for the wrong reason, still green. The control that convinced me it is real is
removing the `cancel()` and leaving everything else alone:

```
with the session cancel      0.02 s
cancel removed, backstop only 5.03 s   (SHUTDOWN_DRAIN)
```

So the open stream genuinely is what blocks the drain, the cancel is what
clears it, and the backstop catches the case where a client ignores the close.
The comment on the assertion names that control so the next person can rerun
it.

The requests are written to the socket by hand. pond has no HTTP client
dependency and this needs to own the socket and decide when it closes, which
is the whole point of the test, so a client crate would have been both a new
dependency and less precise about the thing under test.

`cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` and
`cargo test` are green on 1.95.0.

Disclosure: I am the founder of Dockhold, which is where I was running pond
when I hit both of these. Neither is specific to it.

## Release note

`pond serve` now stops when a supervisor asks it to. It handles SIGTERM as well
as ctrl-c, and shutdown no longer hangs when an agent is connected: live MCP
sessions are closed and the drain is bounded, so restarts finish instead of
waiting for the process to be killed.

### DIFF
--- a/packages/pond/src/transport.rs
+++ b/packages/pond/src/transport.rs
@@ -27,7 +27,11 @@ pub mod http {
     //! `POST /v1/get-message`, and the `/mcp` route carrying rmcp's
     //! streamable-HTTP MCP transport.
 
-    use std::net::{IpAddr, SocketAddr};
+    use std::{
+        future::Future,
+        net::{IpAddr, SocketAddr},
+        time::Duration,
+    };
 
     use anyhow::Context;
     use axum::{
@@ -41,6 +45,7 @@ pub mod http {
         StreamableHttpServerConfig, StreamableHttpService, session::local::LocalSessionManager,
     };
     use tokio::net::TcpListener;
+    use tokio_util::sync::CancellationToken;
 
     use super::AppState;
     use crate::{
@@ -69,11 +74,23 @@ pub mod http {
         hosts
     }
 
+    /// How long shutdown waits after the MCP sessions have been cancelled before
+    /// the process exits regardless. Cancelling the token closes the live
+    /// streams, so the drain normally finishes far inside this; the deadline is
+    /// only a backstop, so that a client which ignores the close cannot hold the
+    /// process open the way an uncancelled stream used to.
+    const SHUTDOWN_DRAIN: Duration = Duration::from_secs(5);
+
     /// Build the axum router: the `/v1/*` JSON handlers plus the nested `/mcp`
     /// streamable-HTTP MCP service. Public so the integration test can drive it
     /// without binding a socket. `allowed_hosts` extends the `/mcp` route's
-    /// loopback `Host` allowlist (see [`mcp_allowed_hosts`]).
-    pub fn router(state: AppState, allowed_hosts: &[String]) -> Router {
+    /// loopback `Host` allowlist (see [`mcp_allowed_hosts`]); cancelling
+    /// `mcp_shutdown` tears down live MCP sessions (see [`serve_with_shutdown`]).
+    pub fn router(
+        state: AppState,
+        allowed_hosts: &[String],
+        mcp_shutdown: CancellationToken,
+    ) -> Router {
         let mcp_state = state.clone();
         let mcp = StreamableHttpService::new(
             move || Ok(super::mcp::PondMcp::new(mcp_state.clone())),
@@ -81,7 +98,8 @@ pub mod http {
             // `with_allowed_hosts` replaces the list, so the helper hands it
             // the defaults plus the extras rather than the extras alone.
             StreamableHttpServerConfig::default()
-                .with_allowed_hosts(mcp_allowed_hosts(allowed_hosts)),
+                .with_allowed_hosts(mcp_allowed_hosts(allowed_hosts))
+                .with_cancellation_token(mcp_shutdown),
         );
         Router::new()
             .route("/v1/search", post(search))
@@ -121,14 +139,82 @@ pub mod http {
             .local_addr()
             .context("failed to read bound address")?;
         tracing::info!(%local, "pond serve listening (HTTP /v1/*, MCP /mcp)");
-        axum::serve(listener, router(state, &allowed_hosts))
-            .with_graceful_shutdown(shutdown_signal())
-            .await
-            .context("axum server error")
+        serve_with_shutdown(listener, state, &allowed_hosts, shutdown_signal()).await
     }
 
+    /// The serving half of [`serve`], with the stop trigger injected. Public so
+    /// the integration test can drive a real socket and a real MCP stream
+    /// without raising a process signal.
+    ///
+    /// Stopping is two-stage, because axum's graceful shutdown waits for every
+    /// in-flight connection and an MCP client holds its `GET /mcp` stream open
+    /// for the life of the session: cancelling the token the streamable-HTTP
+    /// service was built with ends those sessions so the drain can finish, and
+    /// [`SHUTDOWN_DRAIN`] then bounds the wait for anything that ignored it.
+    pub async fn serve_with_shutdown(
+        listener: TcpListener,
+        state: AppState,
+        allowed_hosts: &[String],
+        stop: impl Future<Output = ()> + Send + 'static,
+    ) -> anyhow::Result<()> {
+        let mcp_shutdown = CancellationToken::new();
+        // Fires when `stop` does, so the drain deadline is measured from the
+        // stop request rather than from startup. An un-cancelled token's
+        // `cancelled()` never resolves, which is exactly what keeps the
+        // deadline arm below inert while the server is running normally.
+        let stopping = CancellationToken::new();
+        let server = axum::serve(listener, router(state, allowed_hosts, mcp_shutdown.clone()))
+            .with_graceful_shutdown({
+                let stopping = stopping.clone();
+                async move {
+                    stop.await;
+                    tracing::info!("shutdown: closing MCP sessions, then draining");
+                    mcp_shutdown.cancel();
+                    stopping.cancel();
+                }
+            });
+        tokio::select! {
+            result = server => result.context("axum server error"),
+            () = async move {
+                stopping.cancelled().await;
+                tokio::time::sleep(SHUTDOWN_DRAIN).await;
+            } => {
+                tracing::warn!(
+                    seconds = SHUTDOWN_DRAIN.as_secs(),
+                    "shutdown: connections still open after the drain window; exiting anyway"
+                );
+                Ok(())
+            }
+        }
+    }
+
+    /// Resolve on the first stop request: ctrl-c everywhere, and SIGTERM too on
+    /// unix. SIGTERM is what a container runtime or a service manager sends,
+    /// and a server running as PID 1 gets no default disposition for a signal
+    /// it has not handled - so without this arm `serve` ignored the stop
+    /// outright and was killed when the supervisor's grace period ran out.
     async fn shutdown_signal() {
-        let _ = tokio::signal::ctrl_c().await;
+        #[cfg(unix)]
+        {
+            use tokio::signal::unix::{SignalKind, signal};
+
+            match signal(SignalKind::terminate()) {
+                Ok(mut terminate) => {
+                    tokio::select! {
+                        _ = tokio::signal::ctrl_c() => {}
+                        _ = terminate.recv() => {}
+                    }
+                }
+                Err(error) => {
+                    tracing::warn!(%error, "no SIGTERM handler; stopping on ctrl-c only");
+                    let _ = tokio::signal::ctrl_c().await;
+                }
+            }
+        }
+        #[cfg(not(unix))]
+        {
+            let _ = tokio::signal::ctrl_c().await;
+        }
     }
 
     async fn search(
--- a/packages/pond/src/transport.rs
+++ b/packages/pond/src/transport.rs
@@ -74,12 +74,22 @@ pub mod http {
         hosts
     }
 
-    /// How long shutdown waits after the MCP sessions have been cancelled before
-    /// the process exits regardless. Cancelling the token closes the live
-    /// streams, so the drain normally finishes far inside this; the deadline is
-    /// only a backstop, so that a client which ignores the close cannot hold the
-    /// process open the way an uncancelled stream used to.
-    const SHUTDOWN_DRAIN: Duration = Duration::from_secs(5);
+    /// How long [`serve_with_shutdown`] waits, after the MCP sessions have been
+    /// cancelled, before returning regardless. Cancelling the token closes the
+    /// live streams, so the drain normally finishes far inside this; the
+    /// deadline is only a backstop, so that a client which ignores the close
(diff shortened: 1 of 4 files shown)


packages/pond/tests/integration/transport_http.rs
```diff
@@ -79,7 +83,123 @@ async fn router() -> anyhow::Result<(TempDir, Arc<Store>, Router)> {
...
+    // SHUTDOWN_DRAIN instead of milliseconds, which is what proves the open
+    // stream is really what blocks the drain.
+    assert!(
+        head.starts_with("HTTP/1.1 200"),
+        "the /mcp stream has to be established for this test to mean anything:\n{head}"
+    );
+
+    stop.send(()).expect("serve task should still be running");
+
+    let served = tokio::time::timeout(Duration::from_secs(30), server)
```
tenequm, line 162: **Required.** This stays green with the fix removed. The only outcome assertions are the 200 head, this 30 s timeout, and `served?`. When the deadline arm wins, `serve_with_shutdown` returns `Ok(())` inside 5 s, so deleting `.with_cancellation_token(mcp_shutdown)` in `transport.rs` leaves the test passing - your own control run measures exactly that (5.03 s). As written it guards the backstop, not the cancel.

Take `Instant::now()` before `stop.send(())` and `assert!(started.elapsed() < SHUTDOWN_DRAIN)` - export the const, or duplicate the 5 s here with a comment naming it.

packages/pond/tests/integration/transport_http.rs
```diff
@@ -79,7 +83,123 @@ async fn router() -> anyhow::Result<(TempDir, Arc<Store>, Router)> {
...
+        ),
+    )
+    .await?;
+    // Guard against this test quietly going vacuous: if the stream were ever
+    // refused, the connection would close and shutdown would finish for the
+    // wrong reason. With the session cancel removed, this test takes the full
+    // SHUTDOWN_DRAIN instead of milliseconds, which is what proves the open
+    // stream is really what blocks the drain.
+    assert!(
+        head.starts_with("HTTP/1.1 200"),
```
tenequm: The PR body says the test asserts a 200 `text/event-stream`; this asserts only the status line. Either assert the `content-type` header too or drop the claim from the description.

packages/pond/src/transport.rs
```diff
@@ -69,19 +74,32 @@ pub mod http {
         hosts
     }
 
+    /// How long shutdown waits after the MCP sessions have been cancelled before
+    /// the process exits regardless. Cancelling the token closes the live
+    /// streams, so the drain normally finishes far inside this; the deadline is
+    /// only a backstop, so that a client which ignores the close cannot hold the
+    /// process open the way an uncancelled stream used to.
+    const SHUTDOWN_DRAIN: Duration = Duration::from_secs(5);
```
tenequm: **Required (doc only).** The drain can cut an in-flight `POST /v1/ingest` between commits: `upsert_session_batch` commits messages + parts first and the sessions row after, so a stop landing in that gap leaves message rows with no session row until the next ingest of that session rewrites it. Each Lance commit is atomic and it self-heals, and SIGTERM is strictly better than the SIGKILL it replaces - but for ctrl-c it is a change: before this PR the drain waited for the ingest to finish.

Decision: keep the bound, do not exempt `/v1/ingest` - exempting would make the drain unbounded in exactly the case that can hold the process, and a hosted read side never ingests. Add one sentence here saying the deadline can truncate an in-flight write and that the store converges on the next ingest.

packages/pond/src/transport.rs
```diff
@@ -69,19 +74,32 @@ pub mod http {
         hosts
     }
 
+    /// How long shutdown waits after the MCP sessions have been cancelled before
+    /// the process exits regardless. Cancelling the token closes the live
+    /// streams, so the drain normally finishes far inside this; the deadline is
```
tenequm: "before the process exits regardless" is not what the deadline bounds. It bounds this function's return; `main` still drops the runtime afterwards, which waits for `spawn_blocking` work already started (an in-progress `--with-sync` file read). Reword to "how long `serve_with_shutdown` waits".

packages/pond/src/transport.rs
```diff
@@ -121,14 +139,82 @@ pub mod http {
...
+        state: AppState,
+        allowed_hosts: &[String],
+        stop: impl Future<Output = ()> + Send + 'static,
+    ) -> anyhow::Result<()> {
+        let mcp_shutdown = CancellationToken::new();
+        // Fires when `stop` does, so the drain deadline is measured from the
+        // stop request rather than from startup. An un-cancelled token's
+        // `cancelled()` never resolves, which is exactly what keeps the
+        // deadline arm below inert while the server is running normally.
+        let stopping = CancellationToken::new();
```
tenequm: The second token is redundant. rmcp never cancels the configured token and installs no `DropGuard` on it - it only ever takes `child_token()` - so `mcp_shutdown.cancelled()` resolves if and only if this function cancels it, the same statement that cancels `stopping`. One clone of `mcp_shutdown` into the select arm gives identical behavior and removes this token and the four-line comment above it.

If you added it against a future rmcp that cancels the parent itself, say so in one line and keep it; otherwise drop it.

packages/pond/tests/integration/transport_http.rs
```diff
@@ -79,7 +83,123 @@ async fn router() -> anyhow::Result<(TempDir, Arc<Store>, Router)> {
...
+/// holds its `GET /mcp` stream open for the whole session, so before the
+/// session cancel this hung until the supervisor killed the process. Driven
+/// over a real socket, because the hang is in the connection drain and a
+/// `oneshot` against the `Router` never binds one.
+#[tokio::test(flavor = "multi_thread")]
+async fn shutdown_completes_while_an_mcp_stream_is_open() -> anyhow::Result<()> {
+    pond::embed::init_enabled(true);
+    let temp = TempDir::new()?;
+    let store = Arc::new(Store::open_local(temp.path()).await?);
+    let state = AppState {
```
tenequm: This `AppState` block is byte-identical to the one in `search_and_get_round_trip` below. Factor an `empty_state(temp: &TempDir)` helper in this file and call it from both tests, the way `transport_mcp.rs::synthetic_state` and `sql.rs` do.

packages/pond/tests/integration/transport_http.rs
```diff
@@ -5,7 +5,7 @@
 //! are thin adapters over the shared wire handlers. The router is driven via
 //! `tower::ServiceExt::oneshot` - no socket bind, no HTTP client dependency.
```
tenequm: Now false: the new test binds a `TcpListener`. Keep the "no HTTP client dependency" half and name the shutdown test as the socket-binding exception and why (the hang is in the connection drain, which `oneshot` never exercises).

packages/pond/tests/integration/transport_http.rs
```diff
@@ -79,7 +83,123 @@ async fn router() -> anyhow::Result<(TempDir, Arc<Store>, Router)> {
...
+            name.eq_ignore_ascii_case("mcp-session-id")
+                .then(|| value.trim().to_owned())
+        })
+        .ok_or_else(|| anyhow::anyhow!("no mcp-session-id in response head:\n{head}"))
+}
+
+/// Write one raw HTTP/1.1 request and read back just the response head. Raw
+/// rather than through a client crate: the point is to own the socket and
+/// decide when it closes, which is what this test is about.
+async fn request(stream: &mut TcpStream, raw: &str) -> anyhow::Result<String> {
```
tenequm, line 219: `request()` reads with no deadline. Both call sites run before the 30 s timeout, and there is no per-test timeout, so a regression that stalls before emitting headers hangs CI at the runner limit instead of failing the test. Wrap the two `request()` calls in `tokio::time::timeout`.

tenequm (changes requested):
Diagnosis and fix are right: rmcp only ever derives `child_token()` from the configured token and never cancels it itself, so cancelling it is exactly the intended teardown hook. Verified locally - fmt, clippy, the four `transport_http::` tests, plus a live SIGTERM (exit 0 in 28 ms, was ignored as PID 1) and SIGINT with a `GET /mcp` stream held open (exit 0 in 22 ms, was an indefinite hang).

Requesting changes for one real gap in the test (it stays green with the fix removed) and a handful of cleanups, each inline. Not requested, for the record: tokio never unregisters the libc SIGTERM handler, so a second SIGTERM during a stuck shutdown is now a no-op where it used to kill the process; supervisors escalate to SIGKILL, so leaving it is fine.

Everything else checked clean: `tokio-util` placement and comment, the test's placement (genuinely cross-module, impossible as a unit test), hand-rolled HTTP (the dev-dep enables `client` only), `stop` cancellation-safety (axum polls it in its own task), release note length and format. Once these are in, this merges and ships with #196 in the same release.

Maziar110:
All done in 14ac0e9.

You're right about the test, and it's worse than a gap I left: I had measured that exact case. I ran the control with the cancel removed, got 5.03 s, and put the number in the description as proof the test was meaningful. It proved the opposite and I read it as a win. SHUTDOWN_DRAIN is pub now, the test times the shutdown and asserts it lands inside it, and unwiring the token fails it:


```
shutdown took 5.001700877s, so the drain deadline returned rather than the MCP
session teardown letting it finish inside SHUTDOWN_DRAIN (5s)
The second token is gone, and no, I had no future-rmcp reason for it. Your reading is the right one.
```

The rest are in. Content-type is asserted rather than the claim dropped. The const now says it bounds this function's return and not the process, and that returning on the deadline can cut an ingest between commits with the next ingest of that session converging it. The duplicate state is empty_state, next to how synthetic_state does it. Module doc no longer claims nothing binds a socket. The response-head read has a deadline.

One thing I did differently: the timeout went inside request() rather than around the two call sites, so anything added later gets it too. Happy to move it if you would rather see it at the call sites.

Left the second-SIGTERM point alone, per your note. fmt, clippy and 509 tests green on 1.95.0.

tenequm (approved):
All eight are in, and the control now bites: with `.with_cancellation_token(mcp_shutdown)` deleted, `shutdown_completes_while_an_mcp_stream_is_open` fails in 5.09 s with your message; with it present, 0.17 s. Re-verified fmt, clippy `--all-targets -D warnings`, and the four `transport_http::` tests on 14ac0e9.

The timeout inside `request()` rather than at the call sites is the better placement - keep it.

One thing this round surfaced that is pond's to fix, not yours: the "next ingest of that session rewrites the missing row" sentence I asked you to add is true of the write path, but `pond sync`'s freshness gate keys on the messages table alone and can skip that re-ingest. That is a pre-existing gap in pond's oracle, now #212; fixing it makes your sentence true as written, so nothing to change here. I trimmed the second-token paragraph from the description so the squash body matches the code.

Thanks for the honest read on the control run - "it proved the opposite and I read it as a win" is exactly the failure mode the elapsed-time assertion now guards against.

PR #213, opened 2026-09-01
fix(sync): commit messages last so a cut flush cannot latch as fresh

### LINKED ISSUE #212: sync freshness gate keys on messages alone, so a session cut between the messages and sessions commits is never re-ingested
Surfaced while reviewing #198, which bounds `pond serve` shutdown and so makes a write cut between the two `upsert_session_batch` commits more likely than before. The gap itself is older than that PR.

## What happens

`Store::upsert_session_batch` commits in two steps: messages and parts first (one `append_batches` per table via `append_filtered`, `sessions.rs:1303-1313`), then the session row (`merge_insert_chunks(.., Table::Sessions, ..)`, `sessions.rs:1315-1317`). A process exit between those two commits leaves the session's messages stored with no session row. Each commit is atomic, so the data is consistent per table; the session is just half-present.

The comment on `SHUTDOWN_DRAIN` and the design intent say the next ingest of that session rewrites the missing row. The write path does that (`existing_sessions` at `sessions.rs:1180` -> `sessions_owned` filter at `:1244` -> merge-insert), but `pond sync` never schedules the re-ingest:

- `append_filtered` collects every kept batch and issues a single `append_batches` (`sessions.rs:944`), so once the messages commit lands, the store's max message timestamp for that session equals the source's last-record timestamp.
- The production freshness oracle is `RowmapOracle` (`sessions.rs:5735-5740`). `session_max_ts` reads `RowMetaSet::lookup_max_ts`, which `rowmap.rs:167-177` aggregates from message rows only. There is no intersection with the sessions table.
- `is_session_fresh` (`adapter/mod.rs:304-313`) skips when `source <= stored`. The half-present session is reported fresh on every run, forever.

`Store::session_last_message_ids` (`sessions.rs:1419-1431`) already implements the right rule and its doc comment names exactly this hazard ("keying on messages alone would report it fresh and orphan the missing row. Intersecting with the sessions id-set forces a re-ingest that heals it"). It has no production callers; only benches and tests reference it. The only heal today is `pond sync --verify` (`NoopOracle`), and `pond serve --with-sync` hardcodes `verify: false` (`main.rs:4067`), so an unattended host never runs it. `main.rs:4787-4791` already describes this as "historical M1 damage ... a frozen watermark that mtime can never re-read past".

## User-visible effect

`pond_get_session` returns not-found for the session (`Store::get_session` needs the session row, `sessions.rs:1368`) while `pond_search` still returns its messages, since the rowmap carries `project` and `source_agent` from the messages table. A flush batches many sessions, so one cut can leave every newly-seen session in that batch in this state.

## How the window opens

- `POST /v1/ingest` at shutdown: bounded by `SHUTDOWN_DRAIN` (5 s) after #198; before it, an in-flight ingest either finished (ctrl-c waited forever) or was killed by the supervisor's SIGKILL.
- `pond serve --with-sync`: `spawn_in_serve_sync` (`main.rs:4023-4040`) is a detached `tokio::spawn` with no join handle and no cancellation token. When `serve` returns, `main` drops the runtime (`main.rs:1231`) and the sync loop is dropped at its next await, anywhere in the write. This is the wide path, and #198 makes it fire promptly on every SIGTERM instead of after the supervisor's grace period.
- Any crash or power loss between the two commits, independent of shutdown handling.

Spec rule this violates: spec.md#session-movement-complete requires that an "already ingested" signal derive from stored data such that a partial flush re-ingests on the next run. The messages-only watermark is derived from stored data, but from the first-committed table rather than the last, so it latches the same false "done" the rule forbids.

## Fix

1. Make the sync oracle require a durable session row: give `RowmapOracle` the sessions id-set (a narrow scan of `sessions.id`, or carried into the rowmap at build time) and return `None` for a session absent from it, so `is_session_fresh` re-ingests it. This is what `session_last_message_ids` already does; either wire that in or port its intersection.
2. Gate the `--with-sync` loop on shutdown: keep the join handle, hand it the same cancellation, and let the in-flight cycle finish (or cancel only between cycles) before the runtime is dropped.
3. A `pond sync` conformance test that writes a session's messages without its row, runs a sync over an unchanged source, and asserts the row appears.
4. While there: `serve`'s doc still reads "Bind and serve until ctrl-c"; #198 made that SIGTERM too.

Once (1) lands, the `SHUTDOWN_DRAIN` comment from #198 is true as written.

tenequm:
## Summary

Closes #212.

`pond sync`'s freshness gate keys on the resident row-meta map, built from the messages table, but `upsert_session_batch` committed messages and parts before the session row. A flush cut between those commits (crash, power loss, SIGTERM to a running sync, the runtime drop under `serve --with-sync`) left a session whose messages the map knew and whose row did not exist; its stored max timestamp already equalled the source's last record, so every later sync called it fresh and the row was never written: searchable messages, `pond_get_session` not-found, until someone ran `--verify`. `session-movement-complete` forbids exactly this latch.

The fix is on the write side. The invariant is "the table the freshness key derives from commits last", not "the session row commits last" (that came from the row-keyed watermark the oracle replaced): `upsert_session_batch` now commits the session row and parts concurrently, then messages. Any cut leaves the key absent or behind, the gate never fires, and the next sync converges the session through the existing idempotent filters. No read-side cost, and it also closes the sibling hole the old order had (messages landed, parts not, gate satisfied, parts missing forever). The spec names the invariant in one sentence so it is not lost a second time; the plan doc gets a decision update.

Not done from #212's list: gating the `--with-sync` loop on shutdown. Waiting for an in-flight cycle would bring back the multi-minute shutdown #198 just removed, and with the order fixed a cut sync costs a re-read on the next tick, not data. Closed as unnecessary.

Also folds in the stale `serve` doc from #198 ("until ctrl-c" now names SIGTERM) and retires three comments and the `--verify` help text that called a full re-read the only heal.

## Cost

Commit count unchanged. A flush that adds no new session goes from one stage (messages and parts concurrently) to two (parts, then messages): one commit round-trip (`write_ms_per_commit` 566 ms on lance 11, 638 ms on lance 10, per the gate rows) on S3 writing ticks only; no-op ticks and local stores pay nothing. The alternative first tried here, a per-sync `sessions.id` scan in the oracle, measured ~4 s on every S3 tick (identical verdicts, 1.79 s -> 5.92 s wall on the real corpus) and is recorded as rejected in `docs/benchmarks/results.md`. The stage cost is recoverable later by overlapping the messages fragment upload with the parts commit through the write chokepoint (Lance two-phase write); not done here. Bench-gate write row to be taken after rebasing onto the lance 11 upgrade so it compares against the v11 baseline.

## Tests

- `sessions::tests::messages_commit_last` (unit, unix): makes the messages table unwritable, runs an ingest, and asserts the flush errors with the session row and parts already committed and no message rows - the order itself. Restores the table and re-ingests to prove convergence. Fails under the old order (the messages failure aborted the `try_join!` before the row was written).
- `sync::half_flushed_session_is_re_ingested_not_skipped` (integration): fabricates the state the new order leaves (row and parts, no messages), asserts the oracle yields no key, syncs, and checks the messages appear with the part count unchanged.

`cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test` green.

## Release note

`pond sync` now commits a session's messages last, so a flush cut by a crash or shutdown is re-ingested on the next run instead of being treated as up to date forever; such sessions were searchable but missing from `pond_get_session`. No action needed.


tenequm:
## Parked as draft - design call pending

State of the branch: rebased onto `main` at `0ab11c7` (lance 11), fmt/clippy/tests green, bench-gate row taken against the real S3 store on lance 11. Reads at parity with the same-day lance-11 row; `write_ms_per_commit` 566 -> 548 ms (the append sweep, untouched by this PR). The one alarming delta, `write_fold_ms` +203%, is not this PR: the fold rounds land on messages manifest version 16 instead of 15 because #214's final form stamps the FTS stemmer with a separate `update_config` commit after the index build, and 16 is where the `cleanup_interval = 16` gate fires `cleanup_old_versions` (4651 ms of the 7390 ms). Same cleanup cadence in production, shifted by one version.

Why it is parked: this fix serializes two commits that used to run concurrently. Zero extra requests, but a grown-only flush now waits one extra commit latency (~550 ms typical on this store), and the tail is additive: two draws from the slow-commit distribution (transport timeouts, 503 retries, OCC retry against another writer) instead of max of two. For a store whose hard constraint is no added S3 round-trip latency on the write path, that is not an acceptable permanent cost for a crash window of under a second per flush.

Alternatives weighed, none clean:
- Read-side intersect with the sessions id-set (the first version of this branch): ~4 s cold per tick, on no-op ticks too. Rejected, measured in `docs/benchmarks/results.md`.
- Local write-ahead intent record (session ids of the in-flight flush, cleared after all commits): zero S3 cost, but introduces host-local state into a design where the store is the single source of truth. Rejected.
- Keep commits concurrent and reconcile row/parts presence off the hot path (inside `pond optimize`, which already scans all three tables, or an explicit repair command): zero hot-path cost, latch becomes detected-and-healed instead of prevented. Viable; not yet verified against the current optimize code.
- This PR (messages commits last): correct by construction, pays the serialization above.

Nothing prevents the latch for free: verifying row and parts durability means reading two more datasets, and on S3 that is round trips. The decision is which cost to carry. Until then #212 stays open; a half-flushed session self-heals if it ever grows again and `pond sync --verify` heals the rest.

Issue #166, opened 2026-08-21
tenants: keep work and personal sessions apart, scope what each agent can recall


tenequm:
## Why

The wire protocol already carries an opaque `namespace` (spec 7.3) and the substrate has a catalog seam for it (3.2 `lance-chokepoints-catalog`), but a personal pond has exactly one namespace and nothing enforces separation. Two concrete needs:

- **Work vs personal.** One person's client sessions must not mix with their personal ones in the same store, and an agent working on a client repo must not be able to recall personal sessions (or the reverse). Today the only answer is two separate stores with two config files and two `serve` processes - it works, but it is undocumented and every surface (schedule, status, MCP registration) assumes one.
- **Per-agent scoping.** A fleet of agents sharing one store (e.g. several assistants behind one `pond serve`) where each agent should see only its own tenant.

## Design (agreed 2026-09-01)

Full write-up, findings, alternatives, and edge rules: [docs/plans/2609-01-scoped-access-tenants-design.md](https://github.com/tenequm/pond/blob/main/docs/plans/2609-01-scoped-access-tenants-design.md).

1. **Token gate + token-to-tenant map on `serve`.** `[serve.tokens.<name>]` (`token_file` / `token_command`, `tenants = [...]`) applied on `/v1/*` and `/mcp`; 401 on unknown or missing token; required for a non-loopback bind; stdio takes `--tenant`. Spec 2.3 keeps "no identity" and drops "no authentication".
2. **Session-level editable `tenant` label, not denormalized.** Typed column on `sessions` or an append-only `session_tenants` sidecar (B vs C, decided after the benchmark below). Reads resolve the visible session set and push `session_id IN (...)` through the existing BTree indexes on `messages` / `parts`; `pond_sql` views get the same semi-join. Assigned at ingest from `[adapters.<x>].tenant` or the token; edited with `pond tenant set` / `pond tenant assign`; child sessions follow the parent.
3. **Fail-closed.** Unassigned sessions are visible only to unscoped callers; scoped responses report in-scope counts only; the wire `namespace` narrows within the token's allowlist, anything outside is `namespace_unknown`. Scoped tokens can ingest (tenant stamped on Session events, mismatches rejected per row).

One store, all visible by default, scoped when tokens are configured. No child Lance namespaces or partitions now (deferred to a physical trigger); no MemWAL needed for live writes; storage credentials (`[creds.<name>].scope`) remain the hard boundary.

## Open

[...]

PR #215, opened 2026-09-01
docs(guides): add a Dockhold worked example to the remote serve guide

Maziar110:
## Summary

Takes you up on the offer in #194: one section at the end of
`guides/remote-serve`, after the platform-neutral body. Nothing above it
changes, and the page still reads as host-agnostic with this appended.

It maps your five steps onto one host and adds nothing conceptual: the
`Dockerfile`, which of the four variables is plain config and which two are
secrets, why `POND_ALLOWED_HOSTS` needs a second pass, and the edge token gate
standing in for "put auth in front".

Two details in it are there because they cost me time rather than because they
are interesting:

- `exec` in the `CMD`. Without it the shell stays PID 1, the stop signal never
  reaches pond, and the clean stop you just documented in #198 quietly turns
  back into a kill. Shell form is still needed to expand the injected `$PORT`,
  so it has to be `CMD exec ...`.
- The hostname does not exist until the app does, so `POND_ALLOWED_HOSTS` can
  only be set on a second pass. In between you get precisely the split your
  page describes: `/v1/search` answering and `/mcp` on 403.

I also stated the trust boundary of the edge gate against myself, since the
page is right that `serve` has no auth of its own and a reader deserves to
know that TLS terminates at the platform before the request reaches pond.

Written in the page's voice, third person, no first-person disclosure baked
into the docs. Prose is ASCII, no em-dashes, and the section is plain markdown
with one table so the MDX compiles as-is. I could not run `pnpm build` here
(no `node_modules` in the checkout), so a rendering check on your side is worth
one look.

On what is verified: I have deployed pond on Dockhold and confirmed the 403 and
the `/v1` split on the real edge, and I verified `POND_ALLOWED_HOSTS` and the
five-second drain against a live S3 store locally. What I have not run is that
exact pair together on the platform, because the `--allowed-host` flag only
ships in the next release. Say the word and I will hold this until after the
release and re-run it end to end first.

Happy to cut anything that reads as too much, or to drop the section entirely
if it has stopped feeling worth the page space.

Disclosure: I am the founder of Dockhold.

### DIFF
--- a/docs/site/src/pages/guides/remote-serve.mdx
+++ b/docs/site/src/pages/guides/remote-serve.mdx
@@ -71,3 +71,50 @@ curl -sS -o /dev/null -w '%{http_code}\n' -X POST https://pond.example.com/mcp -
 ```
 
 The first returns a search envelope; the second prints `200`. A `403` on the second with a `200` on the first means the hostname is missing from `POND_ALLOWED_HOSTS`.
+
+## Worked example: Dockhold
+
+[Dockhold](https://dockhold.eu) builds the `Dockerfile` in a repo and puts it on an always-on HTTPS URL, with no servers to manage and hosting in the EU. Always-on is the part that earns its place here: the warm row map and the reused S3 client are exactly what make one long-lived `serve` faster than many cold ones, and a host that sleeps an idle app throws both away. Worth checking against whatever host you pick.
+
+The whole repo is one `Dockerfile`:
+
+```dockerfile
+FROM debian:bookworm-slim
+ARG POND_VERSION=latest
+RUN set -eux; \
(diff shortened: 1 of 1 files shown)


tenequm (approved):
Built the site at your head - section and table render, ASCII clean, so the check you could not run is done. The `exec` note is the kind of detail this page exists for; without it #198 would have been for nothing on this host.

No hold needed: the guide already documents POND_ALLOWED_HOSTS ahead of the release, and the docs commit rides the same release PR. Merging.

Your table also answers the question I left on #166: Dockhold hands secrets as env vars, so the token shape there gets an env mirror. Thanks.

Issue #61, opened 2026-06-19
memory: reduce idle RAM of `pond mcp` toward <500 MiB (stream rowmap build, f16 model)


tenequm:
## Goal

Bring idle RAM of a long-lived `pond mcp` server toward the **<500 MiB** target. Peak stays comfortably under the 2 GiB ceiling.

## Measured baseline (real 2.1M-message corpus, local FS)

`cargo bench --bench serve_mem_bench` after serving fts + vector + sql + get, model idle-unloaded:

- **idle floor: 877 MiB phys_footprint** (the macOS "Memory"/Jetsam metric)

`vmmap` anatomy of that floor:

| Component | MiB | Reclaimable? |
|---|---|---|
| Live Lance caches (index IVF/PQ + FTS postings) | ~200 | no - needed for search |
| `MALLOC_LARGE (empty)` - freed, retained by allocator | ~351 | soft (kernel reclaims under pressure) |
| `MALLOC_SMALL (empty)` - freed, retained by allocator | ~285 | soft |
| base process / stacks / __DATA | ~40 | no |
| `.rmm` resident meta map (mmap) | **0 dirty** | file-backed, not in footprint |

**Key finding:** ~636 MiB of the floor is freed-but-retained transient memory, not live data. The system allocator's high-water mark is set by the largest transient peaks, and it holds those regions. Only ~200 MiB (the Lance caches) is genuinely live - already under target.

## Dead end - do NOT retry: swapping the global allocator

Measured on the same corpus; all WORSE than the system allocator:

| Allocator | idle floor PF |
|---|---|
| system libmalloc (current) | **877** |
| mimalloc (default) | 1525 |
| mimalloc `MIMALLOC_PURGE_DELAY=0` | 1042 |
| jemalloc `dirty_decay_ms:0,muzzy_decay_ms:0` | 1756 |

macOS `phys_footprint` is tightly coupled to the system libmalloc's `MADV_FREE_REUSABLE` path. Third-party allocators reserve/retain arenas that count against the footprint and return poorly on Darwin (jemalloc has long-standing poor macOS support). `malloc_zone_pressure_relief` also does not help (it MADV_FREEs lazily; pages stay in the footprint until kernel pressure). This avenue is closed - the fix must reduce our own transient peaks.

## Proposed work (the real levers)

The retention tracks the high-water mark of two transient peaks:

### 1. Stream the rowmap build (biggest single lever)
[...]

PR #249, opened 2026-09-16
fix(memory): trim retained glibc pages at quiescent points

tenequm:
## Problem

Long-lived Linux `pond mcp` and `pond serve` processes retain freed glibc allocator pages after transient rowmap, sync, and query peaks. Issue [#61](https://github.com/tenequm/pond/issues/61) measured about 636 MiB of freed-but-retained memory in the idle floor, and [#245](https://github.com/tenequm/pond/issues/245) records query-driven floor growth and multi-GB peaks. Allocator replacement is deliberately out of scope because the measured macOS footprint regressed under jemalloc and mimalloc.

## Fix

Add a glibc-only `malloc_trim(0)` helper using the existing `libc` dependency, with a no-op implementation everywhere else and one narrowly scoped unsafe call. The gate is `all(target_os = "linux", target_env = "gnu")` rather than bare `target_os = "linux"`: `malloc_trim` is a glibc extension that the `libc` crate declares only for the gnu environment, so a musl target would not compile against the FFI branch.

The helper runs only at quiescent boundaries:

- after a successful sync pipeline has completed import, rowmap refresh, index maintenance, and summary work;
- after this process finishes a rowmap base build, delta build, or delta compaction - not when it merely opens a map a sibling published, and not when it loses the build lock and falls back to a trailing oracle, because neither built anything;
- once after server startup finishes prewarming, which is this process's largest transient and which a server nobody queries would otherwise hold forever;
- from the existing 30-second serve/MCP rowmap refresh task, at most once per interval and only when a request completed since the prior periodic trim.

No request path ever trims. HTTP and MCP handlers only arm a shared completion flag that the periodic task consumes, so the ceiling under sustained load is one arena walk per refresh interval, on a background task rather than on a request thread; a server nobody is querying pays nothing after its startup trim.

`MALLOC_ARENA_MAX` is deliberately not set here. The generated `pond-sync.service` is one-shot and exits after each run, so it has no retention problem to cap, and the bench evidence puts arena capping second-order behind the trims. It stays a documented deployment knob for custom long-lived `pond serve` units (recommended cap: the effective tokio worker count), not a value pond injects into a user's environment.

Each Linux trim emits a debug event carrying whether glibc reports that pages were released - "nothing to release" is the normal steady-state answer, so it is logged rather than acted on. This does not replace the global allocator and does not change macOS or Windows behavior.

## Verification

- `cargo fmt --check`
- `cargo clippy --locked --workspace --all-targets -- -D warnings`
- `cargo test --locked` (full suite)

RSS behavior is intentionally not asserted in unit tests. What is asserted is the gate the periodic trim depends on: `completed_requests_arm_the_periodic_allocator_trim` (HTTP) proves a completed request - including one answered with an error - arms the flag and that taking it disarms it, so an idle interval cannot trim; the MCP tool round-trip asserts the same flag after its tool calls. Both would fail if a handler lost its activity guard, which is otherwise invisible. The before/after proof belongs to the instrumentation scenarios defined by the [memory instrumentation plan](https://github.com/tenequm/pond/blob/main/docs/plans/2609-16-memory-instrumentation.md).

## Merge gate: satisfied

The B4 merge gate required before/after retained-bytes rows for the `mem_bench` scenarios `mcp-query-growth` and `rowmap-build-cold`. Those rows are posted in the gate comment on this PR and the verdict is pass: at profile `large`, end anon RSS falls about 80% on every sync/rowmap scenario (`rowmap-build-cold` 308.6 -> 62.4 MiB, `sync-noop-local` 301.8 -> 59.9 MiB), and `mcp-query-growth` also loses a peak, 577.2 -> 346.7 MiB (1.66x), because the trim fires between the loop's iterations and therefore inside the measured region. Peak heap is unchanged throughout, which is the expected shape for retained-page reclaim rather than for a leak fix.

## Profiling attribution (2026-09-16)

Large-profile heaptrack/dhat runs on the instrumentation branch reframe what this fix does and does not do:

- The `mcp-query-growth` "retention gap" (579.5 MiB RSS over 57.4 MiB loop heap at the large profile) is residue of the pre-loop rowmap cold build (a 497 MiB, 6.1M-allocation transient), not query-loop retention. The query loop's live heap is flat (about 5 bytes/iteration over 45 iterations): there is no leak.
- This PR is therefore a mitigation, not a cure: `malloc_trim` reclaims the retained small/mid-chunk pool after peaks (the >32 MiB allocations are glibc-mmap'd and already return on free), but nothing here lowers the transient peak itself. Peak reduction belongs to the source fixes (streaming rowmap build, ingest allocation hygiene - follow-ups under #245).
- `MALLOC_ARENA_MAX` is second-order in the single-sequence bench but stays justified for deployments from the field evidence in #61 (concurrent serve traffic across tokio workers). Recommended cap: the effective tokio worker count.

Full analyses: mem-gate rows in PR #253; attribution reports archived by the orchestrator (2609-16-analysis-heaptrack-mcp-query-growth.md, 2609-16-profiling-synthesis-b1-b5-mapping.md).

Refs #245, #61.

## Release note

Linux serve, MCP, sync, and rowmap workloads now ask glibc to release freed pages after transient peaks, reducing retained idle memory without changing the allocator or non-Linux behavior.


tenequm:

**Method.** Before-rows were measured fresh on `feat/memory-instrumentation` @ 44886ce; the rows committed in `docs/benchmarks/mem-gate-baseline.jsonl` predate 6f3e521, which changed the bench's measured region, so they are not a valid comparator. After-rows come from a throwaway local merge of this branch with 44886ce (never pushed), both profiles run strictly sequentially on one host with pre-cached corpora. Peak heap is effectively deterministic here (spread below 0.01% across branches that do not touch a scenario), while peak RSS carries 2-6% run-to-run noise - treat peak-RSS deltas under roughly 6% as not meaningful.

Peaks:

| `mcp-query-growth` | large | 577.2 -> 346.7 | **-39.9% (1.66x)** | 58.1 -> 59.6 | +2.6% |
| `mcp-query-growth` | ci | 191.9 -> 174.1 | -9.3% (1.10x) | 17.3 -> 15.9 | -8.1% |
| `sync-noop-local` | large | 654.8 -> 675.7 | +3.2% | 495.8 -> 495.8 | 0.0% |
| `sync-incremental` | large | 678.8 -> 671.6 | -1.0% | 495.8 -> 495.8 | 0.0% |
| `rowmap-build-cold` | large | 662.8 -> 678.2 | +2.3% | 495.8 -> 495.8 | 0.0% |

Idle floor at end of run, profile `large`:

| scenario | end RSS before -> after (MiB) | delta | end anon RSS before -> after (MiB) | delta |
| --- | ---: | ---: | ---: | ---: |
| `sync-noop-local` | 374.9 -> 132.4 | -65% | 301.8 -> 59.9 | -80% |
| `sync-incremental` | 426.0 -> 161.8 | -62% | 325.4 -> 61.2 | -81% |
| `rowmap-build-cold` | 381.1 -> 135.2 | -65% | 308.6 -> 62.4 | -80% |
| `mcp-query-growth` | 504.7 -> 274.1 | -46% | 391.1 -> 162.0 | -59% |

**Verdict: pass, stronger than expected.** End anon RSS drops roughly 80% on every sync/rowmap scenario at `large`. It also cuts a *peak*: `mcp-query-growth`/`large` goes 577.2 -> 346.7 MiB (1.66x), because that scenario is a 20-iteration loop and the quiescent-point trim fires between iterations, inside the measured region rather than after it. Peak heap is unchanged throughout, the expected shape for retained-page reclaim. The PR description understates the effect.

One note for whoever rebases: merging 44886ce into this branch hits a textual conflict in `packages/pond/src/lib.rs` - both sides insert a `pub mod` line immediately after `pub mod handlers;` (this branch adds `pub mod memory;`, the instrumentation branch adds the `memprobe` module behind a cfg). The rows above were measured on a local merge with that resolved by keeping both lines; there is no semantic overlap, so it is a one-line fix.

PR #252, opened 2026-09-16
fix(sync): bound ingest flush memory

### LINKED ISSUE #229: sync: peak memory grows with the flush batch's bytes, which nothing bounds
Peak RSS during `pond sync` grows linearly with the bytes of the current flush batch, with no byte bound anywhere on the path. A single large session, or 100 ordinary ones, can push a sync into multi-GB territory on a machine that has no reason to need it.

## Measured

Synthetic agy conversations, one 2 KB `step_payload` per step, ingested into a fresh local store with the release binary (peak RSS via `getrusage(RUSAGE_CHILDREN)`, so no sampling race):

| Input | On disk | Peak RSS |
|---|---|---|
| no conversations (floor) | 0 | 106 MB |
| 1 x 2,500 steps | 10.3 MB | 198 MB |
| 1 x 5,000 steps | 20.5 MB | 270 MB |
| 1 x 10,000 steps | 41.1 MB | 411 MB |
| 1 x 20,000 steps | 82.1 MB | 696 MB |
| 10 x 1,000 steps | 41.3 MB | 407 MB |
| 100 x 100 steps | 43.4 MB | 406 MB |

The last three matter: the same 41 MB split across 1, 10 or 100 sessions peaks the same, because all of it lands in one flush. Fitted model: `peak ~= 110 MB + 7 KB x messages + ~10x payload bytes`.

## Where the bytes are

Massif on the 5,000-step run (heap peak 167 MB, RSS 289 MB) attributes it as:

- **~110 MB non-heap floor** - touched pages of the 372 MB unstripped binary, ~30 threads, Lance/DataFusion init. Store *creation* costs ~48 MB more than opening one. Not proportional to input.
- **~60 MB whole-session buffer** - `IngestValidator.messages: Vec<BufferedMessage>` (`packages/pond/src/sessions.rs:3917`, pushed at `:4236`) holds every message of the in-flight session until the session closes. Of that: 14.2 MB base64 raw records, ~22 MB of serde_json BTreeMap nodes for the options tree (a leaf node is a fixed ~630 B block regardless of key count, so ~4.4 KB/msg of structure for ~150 B of information), ~10 MB duplicated `content`, ~9.5 MB of per-message `options.pond` host-stamp clones (`sessions.rs:4138-4143`).
- **~101 MB flush-side materialization** - `upsert_session_batch` -> `messages_batches` (`sessions.rs:5509-5545`) encodes **all** rows before appending any: 33.2 MB of jsonb `Vec<Vec<u8>>` (`:5514-5517`), 47.1 MB Arrow `LargeBinaryArray` for `options` (`:5598-5600`, MutableBuffer doubling means up to 2x slack), 16.8 MB `StringArray` for content/search_text, 3.8 MB of null `vector` column. `chunk_ranges` (`:5388`) exists but its budget is `COLUMN_BYTE_BUDGET = 1 << 30` (`:5382`) - an i32-offset guard, not a memory bound.

## Why there is no bound today

`ingest_adapter` flushes when `pending_substreams() >= ADAPTER_FLUSH_BATCH` - 100 **completed sessions** (`packages/pond/src/handlers.rs:184`, `:368`). Nothing counts bytes, and the in-flight session is never flushed early (`sessions.rs:3990-4002`, `4070-4087`). So the batch is bounded by session *count* while its cost is bytes.

## Suggested direction

A byte budget on the buffered batch (~32-64 MB), flushing the in-flight session's messages when it is exceeded, plus chunked encode/append in `messages_batches` rather than materializing every row first. The write order already tolerates it: messages and parts are appended before the sessions merge (`sessions.rs:1303-1318`) and `append_filtered` drops already-present primary keys (`sessions.rs:925-946`), so a partial flush is idempotent on re-sync. The accounting that assumes one outcome batch per session (`handlers.rs:226-230`, `:482`) is what needs adjusting.

## Scope

Not adapter-specific: the adapter holds nothing at the peak, and the shape being buffered (`options.source.raw_record`) is the seam's documented lossless-projection convention (`adapter/mod.rs:631-635`), which every adapter emits. Found while benchmarking the agy adapter (#225); the agy side was fixed separately by streaming its source rows, which moved the 20,000-step peak only 742 -> 696 MB, confirming the remainder is here.

tenequm:
## Problem

`pond sync` bounded a flush batch by completed-session count but not by bytes. Large messages or a single large in-flight session could therefore make buffered flush memory grow linearly with the batch payload, and message encoding materialized the full batch before any append.

## Fix

- Cap buffered canonical event bytes at 32 MiB and flush completed message prefixes from an in-flight session while deferring its freshness-bearing session row until close.
- Enforce the byte-budget flush in adapter sync, wire ingest, and the production-mirroring benchmark helper.
- Encode and append messages in byte-bounded chunks while preserving append-plus-filter routing and partial-flush idempotency.
- Keep session/message/part outcome counts correct across partial and final flushes.
- Count the fixed per-row vector width in that budget. `embedding_columns` materializes a dense `embedding_dim()` x f16 slot plus an `embedding_model` offset for every message row even when every vector is null (the embedder-less ingest case, 768 + 4 bytes per row), which source JSON bytes cannot see. Both the validator's buffered-byte accounting and the message chunk loop now add it, so that buffer is bounded by the flush window instead of the whole batch. This is the term [#256](https://github.com/tenequm/pond/pull/256) explicitly deferred here (155.6 MiB in one allocation at 200k messages).

## What this bounds, and what it does not

Bounded: the bytes a single flush window buffers and encodes - canonical event JSON plus the fixed per-row vector and embedding-model slots.

Not bounded by this PR: the Lance encoder's own internals (a roughly 147 MiB floor on the ingest scenario), cross-flush dedupe sets, and the rowmap build transient that sets the high-water mark for every sync scenario. Those are separate terms - the rowmap build in [#255](https://github.com/tenequm/pond/pull/255), the per-message `options.pond` stamp clone in [#256](https://github.com/tenequm/pond/pull/256), and allocator page retention in [#249](https://github.com/tenequm/pond/pull/249). Note also that `json_size` under-counts the stamp term by roughly 25x, so the budget's byte proxy is approximate by construction; #256 removes that term rather than measuring it better.

## Measured effect

Three interleaved runs per side on one Linux host, `ingest-large-session --profile large` (200k messages in one session), release build with `mem-probe`, both binaries built from `8c20b7c` and `8c20b7c` plus this PR. Fragments are the `.lance` files under each table's `data/` directory in the store the scenario writes.

| side | wall (ms) | peak heap (MiB) | peak RSS (MiB) | `messages` fragments | `parts` fragments | store bytes (MB) |
| --- | --- | --- | --- | --- | --- | --- |
| main | 3838-4844 | 1368.9-1381.5 | 1502.6-1511.1 | 1 | 1 | 35.11-35.21 |
| this PR, as the gate scenario drives it | 3788-4466 | 1113.2-1124.6 | 1362.4-1369.1 | 7 | 1 | 35.01-35.13 |
| this PR, with the byte-budget flush wired into the harness | 3969-4589 | 338.3-340.8 | 597.3-608.1 | 9 | 9 | 35.01-35.06 |

Two things follow. First, on write amplification: total on-disk data bytes are unchanged to within run noise, so the smaller flush window costs fragments and commits, not bytes - `messages` goes 1 -> 9 files and 3 -> 11 versions, `parts` the same, which `pond optimize` compacts on its normal schedule. Wall time is indistinguishable from main at this size.

Second, the gate scenario understates this PR by about 4x. `benches/mem_bench.rs::ingest_batched` never calls `byte_budget_reached()`: it landed on main in [#253](https://github.com/tenequm/pond/pull/253) in parallel with this branch, so it missed the budget check this branch added to the identical helper in `benches/write_bench.rs`. The gate row therefore measures only the chunked encode - the middle row, 1381 -> 1124 MiB peak heap. The cadence `ingest_adapter` actually drives in production is the bottom row: 1381 -> 340 MiB peak heap and 1507 -> 602 MiB peak RSS. That row came from a local one-line patch to the bench harness, never to the product, and nothing was appended to `docs/benchmarks/mem-gate-baseline.jsonl`. Giving `mem_bench` the check `write_bench` already has is worth a follow-up rather than a change here, since it resets that scenario's recorded baseline.

The other gate scenarios are unmoved, as they should be: `sync-noop-local`, `sync-incremental` and `rowmap-build-cold` all report an identical 495.8 MiB peak heap on `large` (full table in the [gate comment](https://github.com/tenequm/pond/pull/252#issuecomment-5699157321)), a shared rowmap-build transient no flush-budget change can touch.

At the `ci` profile (20k messages) fragment counts are identical on both sides and peak heap is flat, as expected: the buffered total never reaches 32 MiB, so neither the flush trigger nor the chunk cut binds. Wall time there is consistently about 10% higher on this branch (353 -> 390 ms mean over three runs each), which is the `json_size` pass the accounting adds per event - roughly 0.6 microseconds per event, paid whether or not the budget ever binds.

An earlier 2.6-3.4x projection for this PR came from a dhat analysis of the pre-6f3e521 bench; it does not hold against the current bench and is superseded by the rows above, which happen to land in the same range for the production path.

## Rebase note

[#256](https://github.com/tenequm/pond/pull/256) and this PR touch the same code for opposite reasons: this PR keeps the per-message `options.pond` stamp insert inside `IngestValidator::push` so the stamped event is what `json_size` accounts for, while #256 deletes the per-message insert entirely in favour of a shared borrow merged in at encode time. Whoever lands second must reconcile them deliberately - a textual merge can silently produce a path that either double-stamps or never stamps, and the byte accounting has to follow whichever representation survives.

## Verification

- `cargo fmt --check`
- `moon run pond:format pond:lint pond:test` (what CI runs), all green
- `cargo clippy --workspace --all-targets --locked` clean
- `cargo test --locked` on the merge with `main` (`2323fdc`): 391 lib + 58 bin + 174 integration tests pass, 2 ignored. `main` alone is 385 lib; the 6 added are this branch's own (`byte_budget_flushes_complete_messages_from_an_in_flight_session`, `resync_after_partial_flush_does_not_duplicate_rows`, `byte_budget_counts_the_fixed_vector_width_per_message`, `flush_discharges_every_buffered_byte`, `messages_batch_from_cuts_on_the_fixed_vector_width`, `partial_flush_counts_a_relabel_once`).
- The fixed vector width is added at two independent sites, and each has a test that fails when its term is dropped: `byte_budget_counts_the_fixed_vector_width_per_message` for the flush trigger, and `messages_batch_from_cuts_on_the_fixed_vector_width` for the chunk cut, whose rows are small enough that their text alone would fit one chunk many times over.
- `flush_discharges_every_buffered_byte` pins the buffered-byte invariant from both directions: a missing subtraction leaves `buffered_bytes` non-zero after `finish`, and a missing addition trips a `debug_assert` in `flush` rather than being absorbed by `saturating_sub`.
- `partial_flush_counts_a_relabel_once` covers the partial flush's relabel guard, which keeps one relabel from being counted at the partial write and again at close.

## Links

- Fixes #229
- Production context: #245
- Plan: [`docs/plans/2609-16-memory-instrumentation.md`](https://github.com/tenequm/pond/blob/main/docs/plans/2609-16-memory-instrumentation.md)

## Release note
`pond sync` and ingest now cap the bytes a flush buffers - source payload plus the fixed per-row embedding vector width - and encode messages in bounded chunks, so ingest flush memory no longer grows with the size of the batch.


tenequm:

| `ingest-large-session` | large | 1500.9 -> 1395.8 | -7.0% (1.08x) | 1369.4 -> 1266.8 | -7.5% (1.08x) |
| `ingest-large-session` | ci | 258.8 -> 257.7 | -0.4% | 147.9 -> 143.7 | -2.9% |
| `sync-incremental` | large | 678.8 -> 677.3 | -0.2% | 495.8 -> 495.8 | 0.0% |
| `sync-incremental` | ci | 148.5 -> 146.4 | -1.4% | 52.0 -> 52.0 | 0.0% |
| `sync-noop-local` | large | 654.8 -> 674.0 | +2.9% | 495.8 -> 495.8 | 0.0% |
| `rowmap-build-cold` | large | 662.8 -> 678.9 | +2.4% | 495.8 -> 495.8 | 0.0% |

**Verdict: honest miss against the projection.** The only movement the gate resolves is 1.08x on `ingest-large-session`/`large`; everything else is inside noise, and the `ci` ingest heap move is inside that scenario's own 2.4% heap spread.

Two structural reasons, both visible in the rows. First, `sync-noop-local`, `sync-incremental` and `rowmap-build-cold` all report an *identical* 495.8 MiB peak heap on `large` (52.0 MiB on `ci`) - a single shared rowmap-build transient sets their high-water mark, so a flush-budget change cannot move their peak by construction. Second, on ingest the flush term is a minority of the 1.5 GiB peak: the stamp-clone term alone accounts for 527.5 MiB of it, which is why #256 moves that scenario 1.54x while this PR moves it 1.08x.

The projected 400-520 MiB post-fix ingest peak (2.6-3.4x) came from a dhat run against the pre-6f3e521 bench and is not what the gate measures; it should be re-derived against the current bench before merge. The change may still be justified on flush-size bounding and crash-robustness grounds - the rows neither support nor refute those - but the 2.6-3.4x peak reduction should not stand as measured fact.

PR #255, opened 2026-09-16
perf(rowmap): stream the row-meta map build

tenequm:
## Summary

- Pre-reserve the full-scan entry vector from the store row count.
- Replace three corpus-sized dictionary sort buffers with incremental ordered sets.
- Stream the serialized blob through a buffered temp file, then seek back to write fixed metadata before sync and atomic rename.
- Pre-reserve the rowmap compaction hash map and appended-delta capacity.

## Memory context

Heaptrack attributed a 497.04 MiB cold-build transient to a 160.0 MiB entry spine, 138.56 MiB of owned strings, a 121.81 MiB serialization blob, 45.78 MiB of dictionary scratch, and 22.89 MiB of records/spans. This lands tiers 1 and 2: it removes the corpus-sized blob and dictionary scratch and avoids entry-vector growth. Tier 3 scan/build chunking is deferred because it requires a broader ownership/API redesign.

The on-disk layout, magic, reader, and temp-file/rename installation semantics are unchanged.

## Verification

- cargo fmt --check
- cargo clippy -- -D warnings
- cargo test -p pond-db --lib rowmap::tests::
- cargo test -p pond-db ensure_rowmap_
- cargo test -p pond-db rowmap_chain_compacts_and_stays_bounded

Refs #245

## Release note
Row-meta map builds now stream their serialized payload and avoid corpus-sized dictionary scratch allocations, reducing peak memory during cold builds and delta compaction without changing the cache format.


tenequm:

| `sync-noop-local` | large | 654.8 -> 569.8 | -13.0% (1.15x) | 495.8 -> 321.8 | **-35.1% (1.54x)** |
| `sync-incremental` | large | 678.8 -> 569.7 | -16.1% (1.19x) | 495.8 -> 321.8 | **-35.1% (1.54x)** |
| `rowmap-build-cold` | large | 662.8 -> 569.5 | -14.1% (1.16x) | 495.8 -> 321.8 | **-35.1% (1.54x)** |
| `sync-noop-local` | ci | 141.5 -> 140.3 | -0.8% | 52.0 -> 37.1 | -28.7% (1.40x) |
| `sync-incremental` | ci | 148.5 -> 151.1 | +1.7% | 52.0 -> 37.3 | -28.4% (1.40x) |
| `rowmap-build-cold` | ci | 148.9 -> 138.8 | -6.8% (1.07x) | 52.0 -> 36.6 | -29.6% (1.42x) |
| `mcp-query-growth` | ci | 191.9 -> 188.0 | -2.0% | 17.3 -> 18.3 | +6.0% |

**Verdict: pass, partial by design.** The three scenarios that share the rowmap-build transient all carried an identical 495.8 MiB peak heap on `large` (52.0 MiB on `ci`), and this is the only branch of the set that moves it: 1.54x at `large`, ~1.40x at `ci`, peak RSS following at 1.15-1.19x. All three land on the same post-fix value to within 0.01 MiB, confirming the transient is one shared cost that streaming shrank rather than shifted.

It does not eliminate it. A 321.8 MiB shared transient remains and still sets the high-water mark for all three scenarios, so this is a ~35% cut of the term, not its removal. The path to the ~110-150 MiB target is the tier-3 chunked scan window tracked in #257; this PR is the first step of that, not the whole of it.

One observation worth recording: `mcp-query-growth`/`ci` peak heap moves 17.3 -> 18.3 MiB (+6.0%). It is small and it sits inside that scenario's own run-to-run heap spread (about 8% across branches that do not touch it), unlike the sync/rowmap scenarios where heap is deterministic - so it is noted, not treated as a regression.

PR #256, opened 2026-09-16
perf(ingest): stop per-message options stamp clones

tenequm:
## Problem

dhat attribution of the `ingest-large-session` bench scenario (200,000 messages through the real batched ingest path) shows a 1357.93 MiB peak of which only 39.3 MiB (2.9%) is actual ingested payload ([#245](https://github.com/tenequm/pond/issues/245)). The dominant avoidable term:

- 491.1 MiB (36.2% of peak): the process-wide `options.pond` stamp - a ~100-byte serde_json value - was deep-cloned PER MESSAGE (~2575 bytes of BTreeMap nodes each; 2.2M of the 5.6M live blocks at peak), with every clone held buffered until flush.

## Fix

The change is confined to `packages/pond/src/sessions.rs`:

- Buffered messages now hold the process-wide stamp as a shared borrow (`Option<&'static Value>` from the existing `OnceLock`) instead of a per-message deep clone. A custom serializer merges the borrowed value into the JSONB bytes at encode time, and skips an existing `pond` key so a duplicate is structurally impossible; incoming `options.pond` is still stripped immediately (anti-spoofing invariant unchanged). A regression test asserts pointer-identical stamps across buffered messages and byte-identical encoded output vs the old clone-and-insert representation.

Not changed: `current_parts` capacity (the ~114 MiB dead-spine term) - a safe fix needs conditional first-part logic beyond this PR's scope; left as an open item.

## Not addressed here

The 155.6 MiB dense all-null embedding buffer stays. An earlier revision of this PR omitted the nullable `vector` and `embedding_model` columns from all-null write batches, but that corrupts stores: pond compaction runs `CompactionMode::TryBinaryCopy`, and lance 11.0.0 stamps the full 11-field dataset schema onto page-copied files that physically hold only the 9 written columns, so post-`pond optimize` reads fail with "projection specified the column index 10 but there are only 9 columns". The `index_fold::fold_survives_compaction` integration test reproduces it. That work is reverted here. The term shrinks under [#252](https://github.com/tenequm/pond/pull/252)'s chunked flush budgeting, which accounts for vector byte width as of merged 6c2d09d (`message_row_fixed_bytes`, charged in both the flush budget and the chunk cut), rather than by omitting columns.

## Rebase (done)

Refreshed onto `main` after [#252](https://github.com/tenequm/pond/pull/252). #252 had moved the per-message stamp insert into `IngestValidator::push` so `json_size` would count it; this PR buffers a shared borrow instead. Reconciled so each message is stamped exactly once and the byte budget still matches what gets encoded:

- `push` now only strips any incoming `options.pond` (anti-spoofing unchanged) and never inserts a stamp; `push_message` records the shared `&'static` stamp, and `message_options_bytes` merges it in once at encode time. The `messages_batch_from` chunk cut measures those encoded bytes, stamp included.
- Since the buffered event no longer contains the stamp, `json_size` alone would under-charge it. `push` adds `ingest_host_stamp_bytes`: `"pond":` + the cached stamp length + a separating comma when other keys remain. That is exactly the difference the old insert-then-measure path produced, so #252's flush trigger fires at the same byte count as before.
- New test `byte_budget_charges_the_host_stamp_once` asserts the charged bytes equal `json_size` of the old stamped representation for empty, spoofed-only, and non-empty options. Mutation-checked: removing the stamp-byte charge fails it (case 0), dropping the comma term fails it (case 2), re-adding the per-message clone insert fails it plus `buffered_messages_share_the_host_stamp_until_encoding` and the encode-time `debug_assert` (double stamp), and a `None` stamp fails `ingest_stamps_host_provenance_on_messages_and_strips_spoofed_pond_key` (never stamped).

## Verification

- `cargo fmt --check`, `cargo clippy --all-targets` clean (deny-all active, zero warnings)
- Full `pond-db` integration suite green, including `index_fold::fold_survives_compaction`
- `pond-db` lib unit tests green
- `cargo test --locked` on the merge with `main` (`03d2a46`): 405 lib + 58 bin + 174 integration pass, 2 ignored. `main` alone is 403 lib; the 2 added are `buffered_messages_share_the_host_stamp_until_encoding` and `byte_budget_charges_the_host_stamp_once`
- `cargo clippy --workspace --all-targets --locked -D warnings` clean, also with `--features mem-probe` and `--features dhat-heap`
- Written store bytes: same rows, same nulls, same schema

Memory before/after gate rows are gathered separately per the [memory instrumentation plan](https://github.com/tenequm/pond/blob/main/docs/plans/2609-16-memory-instrumentation.md); expected effect is on the `ingest-large-session` scenario peak (1494.4 MiB RSS / 1370.4 MiB heap at the large profile).

Refs #245.

## Release note
Ingesting large sessions now uses substantially less memory: the shared metadata stamp is no longer copied per message. Stored data is unchanged.


tenequm:

| scenario | profile | peak RSS before -> after (MiB) | delta | peak heap before -> after (MiB) | delta | wall (ms) |
| --- | --- | ---: | ---: | ---: | ---: | ---: |
| `ingest-large-session` | large | 1500.9 -> 973.4 | **-35.1% (1.54x)** | 1369.4 -> 882.7 | **-35.5% (1.55x)** | 3751 -> 2652 (-29%) |
| `ingest-large-session` | ci | 258.8 -> 204.3 | -21.0% (1.27x) | 147.9 -> 99.9 | -32.5% (1.48x) | 338 -> 289 (-15%) |

**Verdict: pass, on target.** The `large` peak drops 527.5 MiB of RSS and 486.7 MiB of heap, which matches the ~491 MiB per-message `options.pond` stamp-clone term identified in profiling plus the allocator overhead that rode on it. The `ci` profile moves in the same direction at a smaller absolute scale. Wall time falls 29% at `large` alongside the memory cut, so removing the per-message clone is a throughput win as well - the copies were never free.

The 155.6 MiB dense all-null embedding buffer intentionally remains in this PR. Omitting it was tried and reverted after the Lance binary-copy compaction finding, so it is left in place rather than worked around here; it is tracked in #257, where the safe lever is the chunk-budgeting rider from #252 rather than a direct change to the column materialization.

No other scenario moves outside noise. `mcp-query-growth`/`ci` shows peak heap +4.8%, which sits inside that scenario's own run-to-run heap spread (about 8% across branches that do not touch it) and is not read as a regression.

PR #262, opened 2026-09-16
build(nix): make the flake the single source of truth for the toolchain

tenequm:
## Summary

Phase 4 of [docs/plans/2609-16-ci-nix-toolchain-revamp.md](docs/plans/2609-16-ci-nix-toolchain-revamp.md): the flake becomes the single source of truth for every tool, and local dev enters it through direnv.

- `flake.nix` gains a `rust-overlay` input and a separate `nixpkgs-toolchain` input. The whole devShell resolves against `nixpkgs-toolchain`, so a routine `nixpkgs` bump (which now only feeds `packages.pond`) moves no compiler store path - cargo does not rerun build scripts and moon/kache keys do not miss.
- Rust comes from `rust-bin.fromRustupToolchainFile ./rust-toolchain.toml` (pure eval, no IFD). `rust-toolchain.toml` stays the one Rust pin. **No rustup in the shell** - it would be a second pin resolving against a different source.
- The shell also carries what the runner image and `bootstrap/action.yml` install today: zig 0.16 (with the `synchronization.def` copy baked in), cargo-zigbuild 0.23.4, rcodesign 0.29.0, gh 2.100.0, the macOS SDK 15.5 stubs behind `SDKROOT`, and fetchurl-pinned moon, protoc, uv, node/npm and kache, plus cmake, pkg-config, libclang and python3.
- `lib.toolVersions` is the one version table, written as plain `name = "x.y.z";` literals so the Windows leg can read it as text. A new flake-check step fails when that text extraction and `nix eval --json .#lib.toolVersions` disagree.
- `.envrc` (`watch_file rust-toolchain.toml` + `use flake`), `/.direnv/` ignored, and AGENTS.md documents `direnv exec . <cmd>` over `nix develop -c`.

Not in this PR (other phases / other PRs): the bootstrap actions are untouched, so the Windows `Select-String` extraction is still deferred, and nothing in CI consumes the devshell yet.

### The `self`-reference constraint holds

The devShell references neither `self` nor the source tree, which is what makes phase 5's profile key sound. Verified rather than asserted: editing an unrelated tracked file (`README.md`) leaves `devShells.x86_64-linux.default.drvPath` byte-identical (`/nix/store/09iw60wyhqz52fv5gygfkh4rwgx2a2sz-pond-dev.drv` before and after), and the derivation's closure contains no flake-source store path.

## Verification (all run on this branch)

**`nix flake check --all-systems --no-update-lock-file`** - `all checks passed!` on aarch64-darwin (19s), all three devShell derivations evaluate. The flake-check job is green on this branch too (39s), with the new parity step passing on ubuntu-latest.

**Devshell tools, aarch64-darwin** (`nix develop --command ...`):

```
rustc 1.95.0 (59807616e 2026-04-14)   cargo 1.95.0 (f2d3ce0bd 2026-03-21)
zig 0.16.0                            cargo-zigbuild 0.23.4
apple-codesign 0.29.0                 moon 2.5.1
libprotoc 35.1                        uv 0.11.33
node v24.18.1 / npm 11.19.0           kache 0.15.0
gh 2.100.0                            cmake 4.4.2 / Python 3.14.7
CARGO_HOME=/Users/tenequm/.cargo-pond
SDKROOT=/nix/store/...-apple-sdk-14.4/...   (the stdenv's own; the 15.5 stubs are Linux-only)
```

**Devshell tools, x86_64-linux** - built and run for real on a Linux box (not just evaluated), since that is where CI runs. Same versions; `SDKROOT=/nix/store/...-macosx-sdk-15.5/MacOSX15.5.sdk`, `cc` resolves to the gcc wrapper (matching the runner image's `build-essential`), and `zig env` reports `lib_dir` in the patched copy with `libc/mingw/lib-common/synchronization.def` present.

**`nix eval --json .#lib.toolVersions`**, verbatim:

```json
{"cargoZigbuild":"0.23.4","gh":"2.100.0","kache":"0.15.0","macosSdk":"15.5","moon":"2.5.1","node":"24.18.1","npm":"11.19.0","protoc":"35.1","rcodesign":"0.29.0","rust":"1.95.0","uv":"0.11.33","zig":"0.16.0"}
```

The new flake-check step's extraction was run locally against this: `diff -u` clean.

**rustc parity (plan checklist item 1): exact, on both platforms.** rust-overlay's rustc and rustup's rustc for the same `rust-toolchain.toml` print byte-identical `-vV`:

| | rust-overlay | rustup |
|---|---|---|
| aarch64-darwin | `rustc 1.95.0 (59807616e 2026-04-14)` / commit-hash `59807616e1fa2540724bfbac14d7976d7e4a3860` / LLVM 22.1.2 | identical |
| x86_64-unknown-linux-gnu | same commit-hash, `host: x86_64-unknown-linux-gnu`, LLVM 22.1.2 | identical |

kache keys on the `-vV` text, not the path, so **no kache prefix rotation is needed** for the rustc component of the key when CI switches to the devshell. (The C-compile path is a separate question - the plan's `KACHE_VERIFY=1` check under the Nix cc wrapper is still open.)

**Darwin `mkShell` vs `mkShellNoCC` (plan checklist item 7): kept `mkShell`.** The concern was the Darwin clang wrapper plus the stdenv's apple-sdk env breaking the `cc` crate or aws-lc-sys. Tested with a real `cargo build --locked --lib` in the shell on aarch64-darwin, from an empty `CARGO_HOME`: it finished green (`Finished dev profile in 22m 03s`). aws-lc-sys 0.41.0 ran its CMake/cc build and emitted `libaws_lc_0_41_0_crypto.a` with `cargo:rustc-link-lib=static=aws_lc_0_41_0_crypto`; ring, blake3, zstd-sys and libsqlite3-sys built their native halves too. No `mkShellNoCC` special-case needed; if one is ever added it must be Darwin-only, since Linux needs the cc wrapper for these same crates.

## Decisions and deviations, flagged

1. **Pins match `main`'s CI today, not the plan's phase-2/3 targets.** `toolVersions` carries moon 2.5.1, protoc 35.1, uv 0.11.33, kache 0.15.0, node 24.18.1/npm 11.19.0 - what `bootstrap/action.yml` and `windows-bootstrap` install on this branch's base - rather than the plan's moon 2.5.5 / kache 0.22.0, which belong to phases 2-3 and have not landed. Matching today's CI keeps the devshell and the runners on one toolchain. #244 bumps the same pins on the bootstrap side (rust 1.98.1, moon 2.5.4, protoc 36.1, uv 0.12.13, kache 0.21.0): whichever of the two merges second must move `toolVersions` in the same commit. Rust needs no edit there - it is read from `rust-toolchain.toml`.
2. **`SDKROOT` is exported on Linux only.** The 15.5 stub tree is the cross-link SDK; on darwin the stdenv's own apple-sdk `SDKROOT` is the correct one and overriding it would break native builds. The SDK package is likewise Linux-only in `packages`.
3. **`mkShell`, not `mkShellNoCC`** - see the verification above.
4. **libclang without a second clang on PATH.** `LIBCLANG_PATH` is set for bindgen, but the C compiler stays the stdenv default (gcc wrapper on Linux, matching the image's `build-essential`; the clang wrapper on darwin). Adding `clang` to `packages` would silently change who compiles `cc`-crate code, and with it every kache key.
5. **python3 is in the shell** though the plan does not list it: `ops/scripts/*.sh` and the dist build's `patch-macos-sdk.py` need it, and today they borrow the runner image's.
6. **npm is installed over node's bundle.** node 24.18.1 ships npm 11.16.0; `.moon/toolchains.yml` pins 11.19.0, so the registry tarball is unpacked over `lib/node_modules/npm`.
7. **The zig patch is a symlink-copy, not a rebuild.** zig's `lib/` tree is copied with `cp -as` (files stay shared), the def file is written into the copy, and the binary is wrapped with `ZIG_LIB_DIR` pointing at it - `zig env` confirms. Rebuilding zig from source to add one file was not worth it.
8. **`CARGO_HOME` defaults to `~/.cargo-pond`, and an explicit `CARGO_HOME` still wins**, so CI's `/ci-cache/cargo` keeps working when phase 5 lands. The shell warns if that directory holds rustup proxies, since moon's rust plugin puts `$CARGO_HOME/bin` first on PATH.
9. **Both nixpkgs inputs currently resolve to the same nixos-unstable rev.** They are separate lock entries on purpose and drift the moment one is bumped alone - which is the point.
10. **The four nixpkgs-sourced versions are asserted, not just declared**: evaluation fails if `nixpkgs-toolchain` moves zig, cargo-zigbuild, rcodesign or gh off the pinned version, so the text table cannot quietly start lying.


tenequm:
Merged `main` in (merge commit `5be78ce`). The only conflict was the `devShells` block in `flake.nix`: #244 added `rustup` to the old three-package shell, this branch replaces that shell wholesale. Kept the flake-owned toolchain and dropped `rustup` - plan 2609-16 §2.1 forbids rustup proxies, because moon's rust plugin prepends `$CARGO_HOME/bin` to PATH and a proxy there would shadow the Nix rustc.

#244's version intent was carried into `lib.toolVersions` instead:

| tool | pin |
|---|---|
| rust | 1.98.1 (read from `rust-toolchain.toml`) |
| moon | 2.5.4 |
| protoc | 36.1 |
| uv | 0.12.13 |
| kache | 0.21.0 |
| node / npm | 24.21.0 / 11.19.1 |

Verified locally: `nix flake check --all-systems` green, the flake-check parity step passes, and in the devshell `rustc --version` is `1.98.1 (48a229cea 2026-09-01)` on `aarch64-apple-darwin` with moon/kache/protoc/uv/node/npm all matching the table. `flake.lock` needed no bump - the pinned rust-overlay already carries 1.98.1.

**Coordination note:** the sibling PR on `feat/ci-moon-kache-refresh` bumps kache to 0.22.0 and moon to 2.5.5. Whichever of the two merges second must bump `lib.toolVersions` (and the matching `sources` hashes here) to match - the flake-check parity step will catch the drift, but the fix is a manual re-pin, not an automatic one.

PR #265, opened 2026-09-16
ci(nix): phase 5 - enter the devshell in CI and delete the bootstrap

tenequm:
Phase 5 of [`docs/plans/2609-16-ci-nix-toolchain-revamp.md`](docs/plans/2609-16-ci-nix-toolchain-revamp.md), repo half. **Stacked on #262** (it contains #262's commits) and must merge after it. Draft until the ops half is deployed.

## What changes

The three `pond-ci` jobs (`build-and-test`, `dist-build`, `publish-release`) replace `- uses: ./.github/actions/bootstrap` with `- uses: ./.github/actions/devshell`, and the bootstrap is deleted.

The new composite action is plan section 2.2's design: hash `flake.nix` + `flake.lock` + `rust-toolchain.toml` into a profile key, `nix print-dev-env --json` against `/nix/var/nix/profiles/pond-dev-$key` (creating it with `--profile` on a miss, which also makes it a GC root), then pipe the result into `$GITHUB_ENV`/`$GITHUB_PATH`. Every later step runs in the shell env with no `nix develop -c` wrapper. Measured locally: a profile hit answers in ~0.1s and skips evaluation entirely, which is the whole point - the flake eval cache is keyed on the git fingerprint, so every commit misses it while the profile path does not.

A second step in the same action verifies the toolchain (`rustc -vV`, `cargo`, `moon`, `protoc`, `kache`, `zig`, `cargo-zigbuild`, `rcodesign`, `gh`, `xz`) so a bad entry fails there instead of ten steps later.

## What the bootstrap did, and where each part went

| Bootstrap did | Now |
|---|---|
| install moon, protoc, uv, kache by curl into `/ci-cache/bin` | from the flake devShell |
| install rustup + `rustup toolchain install` | gone; the devshell ships the pinned rustc |
| `flock` on `/ci-cache/.bootstrap.lock` | gone; nix locks its own store and profile generations |
| `CARGO_HOME`, `CARGO_TARGET_DIR`, `npm_config_cache`, `UV_CACHE_DIR` | **preserved**, exported by the new action |
| `PROTO_HOME=/ci-cache/proto` | **preserved** - see the decision below |
| `RUSTUP_HOME=/ci-cache/rustup` | dropped with rustup |
| `KACHE_CONFIG`, `KACHE_RUNTIME_DIR` | **preserved**, exported by the new action |
| kache prefix-vs-env assertion | **preserved**, moved to a step in `dist-build` (the only job with a kache remote) |
| `mkdir -p` the cache dirs | **preserved** |
| `/ci-cache/bin` and `/ci-cache/cargo/bin` on `$GITHUB_PATH` | dropped - putting `/ci-cache/cargo/bin` on PATH is exactly the rustup-proxy shadowing the plan warns about |

## Decisions where the plan was silent

1. **Composite action, not three inline copies.** Still "one step at the top of each `pond-ci` job", and the reasoning stays in one place.
2. **Cache env lives in the action, not in workflow-level `env:`.** The plan says "workflow-level"; that cannot be right - `CARGO_TARGET_DIR=/ci-cache/target` on `ubuntu-latest` would point `release-prep`, `npm-publish` and `winget-publish` at a path that does not exist there.
3. **`PROTO_HOME` stays,** against 2.2's "PROTO_HOME gone". That removal assumed the versionless `.moon/toolchains.yml` from phase 3, and node/npm are still pinned there (deliberately, per #264's last commit), so moon still provisions them through proto. Without `PROTO_HOME` on the PVC every ephemeral pod re-downloads a node toolchain. It goes when the pins go.
4. **`rust-toolchain.toml` carries all three cross targets** (`aarch64-apple-darwin`, `aarch64-unknown-linux-gnu`, `x86_64-unknown-linux-gnu` - matching exactly what `repo:build-dist`'s `cargo zigbuild` invocations target), and `rustup target add` leaves `moon.yml`. Verified concretely: the `x86_64-linux` devShell's toolchain derivation already references `rust-std` for all three, so the old "minimal profile does not reliably fetch rust-std" caveat does not apply to rust-overlay. Side effect, documented in `windows-bootstrap`: the Windows leg's rustup now downloads those three too (~90 MB it never links against). One pin file is worth it.
5. **release-plz keeps the action's own binstall.** `publish-release`'s `rustup toolchain install stable` step and `RUSTUP_TOOLCHAIN: stable` are deleted rather than replaced - with no rustup, the step is a missing command and plain cargo ignores the variable. The action's normal path pulls a prebuilt release-plz and needs no Rust at all. Its degraded path (GitHub GraphQL down, so binstall cannot find the asset) now source-builds on pond's pinned rustc rather than stable, which may be too old: that is a loud failure on a step `gh run rerun --failed` retries, not a silent second toolchain. `release-prep`, which runs on every main push and is not on `pond-ci`, keeps its hosted `dtolnay/rust-toolchain@stable`.
6. **jq is added to the devshell.** `cargo metadata | jq` in `dist-build` runs on `pond-ci` today and the stripped image guarantees nothing. The entry step itself needs jq *before* the shell is on PATH, so it falls back to `nix run --inputs-from . nixpkgs#jq` - resolved from the flake's own lock, not the registry.
7. **The `/ci-cache` cleanup is enforced, not assumed.** The verify step fails the job if `$CARGO_HOME/bin/{rustc,cargo,rustup}` still exists, with the remedy in the error message. Plan 2.2 calls that cleanup required, and a leftover proxy silently shadows the Nix rustc through moon's rust plugin - a silent wrong-compiler release is worth a loud red.
8. **Binary cache: read only.** `extra-substituters` / `extra-trusted-public-keys` go in the flake's `nixConfig` *and* as explicit flags on the CI nix invocation, because nix ignores a flake's config for a user it does not trust. `flake-check` gets them through `install-nix-action`'s `extra_nix_config`. **Pushing is not wired** - plan 2.8 point 2 puts the signing/push credential on main-branch jobs only, and I cannot create secrets. See pre-merge requirements.

## clang/libclang verdict: removed

`flake.nix` set `LIBCLANG_PATH` but shipped no clang. `bindgen` does not appear in `Cargo.lock` at all, so nothing in the resolved graph can run it: the `-sys` crates that *can* use it resolve to their prebuilt-bindings path, and `aws-lc-sys 0.41.0` depends only on `cc`, `cmake`, `dunce`, `fs_extra`. The variable pointed at a closure nothing used, so it and the plan's "keep clang/libclang (bindgen)" line are both dropped. The comment in `flake.nix` records what to re-add (both together, not the env var alone) if a bindgen consumer ever lands.

## Other cleanups in scope

- The stale comments claiming the runner image ships nix-bin "only for `nix-hash`" are gone (nothing invokes `nix-hash`). `flake-check` stays on `ubuntu-latest` for the reason that is still true: `pond-ci` pods run nix with `sandbox = false`.
- `pkgs.xz` is declared rather than inherited from `stdenv`'s `initialPath` - `moon.yml`'s `tar -cJf` compresses the released artifact.
- Comments in `windows-bootstrap`, `.github/kache/linux-dist.toml` and `.moon/toolchains.yml` that pointed at the deleted bootstrap now point at `flake.nix`'s `toolVersions` or the new action.
- `.moon/toolchains.yml` pins are unchanged, on purpose.

## CI expectations on this PR - confirmed by the first run

`flake-check` (ubuntu-latest) and `windows-verify` **pass**. `windows-verify` passing is the check that the three cross targets added to `rust-toolchain.toml` do not break the leg that reads that file with rustup.

`build-and-test` **fails on `pond-ci`, and the log says exactly why**:

```
warning: '/nix/var/nix' does not exist, so Nix will use
         '/home/runner/.local/share/nix/root' as a chroot store
...
error: executing '/nix/store/svx59425zxp552p2b8gm11qj5r09b56i-bash-5.3p15/bin/bash':
       No such file or directory
```

There is no `/nix` store volume on the current runners, so the static nix falls back to a chroot store under `$HOME`, substitutes a few hundred paths into it, and then dies on the first real build because a `/nix/store/...` path is only valid *inside* that chroot. This is the ops half of phase 5, not something in this diff.

I did not work around it. I did make it fail fast and legibly: the action now asserts `/nix/var/nix` exists and `/nix/store` is writable before it spends 37s downloading into a store it cannot build in, and names the missing infra in the error.

The same run also confirms the reasoning behind decision 8 - `warning: ignoring untrusted flake configuration setting 'extra-substituters'` - which is why the substituter is passed as an explicit flag as well as declared in `nixConfig`.

Once the image and PVC land, expect the first `pond-ci` run to be slow regardless: the `/nix` store starts empty and `/ci-cache`'s warm rustup/proto tree is no longer what these jobs read.

## Pre-merge requirements

1. **Deploy `ghcr.io/tenequm/pond-runner:10`** (stock actions-runner + static nix 2.35.2 at `/usr/local/bin/nix`, `sandbox = false`, local store, `/nix` owned by 1001) to the `pond-ci` scale set.
2. **Attach the `/nix` store volume** to the runner node pool, with the `nix.conf` from plan 2.2 (`http-connections = 50`, `max-substitution-jobs = 32`, `download-buffer-size = 268435456`, `narinfo-cache-negative-ttl = 0`).
3. **One-time `/ci-cache` cleanup:** delete `/ci-cache/rustup`, `/ci-cache/bin`, and the rustup proxies in `/ci-cache/cargo/bin` (**keep the registry cache**). Note the deviation from plan 2.2: `/ci-cache/proto` is **not** part of this cleanup, because `PROTO_HOME` survives (decision 3). The new action fails the job with this remedy if the proxies are still there.
4. **Populate the binary cache** with the toolchain closure for `x86_64-linux` before the first run, or the first cold job pays a full toolchain build. The bucket and keys are managed on the ops side.
5. **Follow-up, not this PR: cache pushing.** Plan 2.8 point 2 wants a signing/push credential on main-branch jobs only. That needs a repo secret (a name like `NIX_CACHE_SIGNING_KEY` plus S3 write credentials) which I cannot create; until then the cache only fills from whatever the ops side pushes.
6. **Fork policy (plan 2.8):** turn on "Require approval for all outside collaborators" in repo settings. Also not something this diff can do.
7. Merge **after** #262.


tenequm:
Infra-side status of the pre-merge requirements (prepared on infra branch `feat/runner-nix-deploy`, applies right after this merges):

1. Runner image :10 - **pushed to ghcr and wired** in the branch's helmfile.
2. /nix volume - **prepared**: 50Gi PVC + `nix-store --init` init container (fsGroup alone leaves a fresh volume failing this PR's precondition guard).
3. One-time /ci-cache cleanup - **prepared** as a run-once Job (deletes rustup dirs + all seven proxies incl. rustfmt/clippy ones, keeps registry + proto).
4. Binary cache - **seeded** (full x86_64-linux devshell closure, 108 paths, signed).
5. Cache pushing - **drafted** as #268 (stacked on this PR); `NIX_CACHE_SIGNING_KEY` secret is set.
6. Fork policy - **done**: repo now requires approval for all outside collaborators' workflow runs.

Also added on the infra side: a weekly `nix-store-reaper` CronJob for the pond-dev-* profile GC roots (atime-based with a noatime guard; a 3-line usage-marker touch in this PR's action would give it a better signal - candidate follow-up).

PR #268, opened 2026-09-16
ci(nix): push the devshell closure to the binary cache

tenequm:
Phase 5 of [`docs/plans/2609-16-ci-nix-toolchain-revamp.md`](docs/plans/2609-16-ci-nix-toolchain-revamp.md): the binary-cache **push** half. #265 (merged) wired the cache for read only; this adds the writer. Draft until phase 5's ops half - the runner image and the `/nix` store volume - is deployed, since nothing here runs before a `pond-ci` job can enter the devshell.

## What changes

- `.github/actions/devshell` gets a `profile` output (`/nix/var/nix/profiles/pond-dev-$key`), written only after the profile exists. The push copies the profile the job actually entered, so it neither re-derives the key nor re-evaluates the flake.
- `build-and-test` gets one last step that `nix copy`s that profile's closure to `s3://pond-nix-cache` (nbg1, zstd), signing each path on upload via the store URI's `secret-key=`.
- Plan doc: phase 5 row and notes updated for the push (a fourth place a cache rotation must edit: the `s3://` URI).

## Gating

```yaml
if: ${{ !cancelled() && github.event_name == 'push' && github.ref == 'refs/heads/main' && steps.devshell.outputs.profile != '' }}
continue-on-error: true
```

- **Main pushes only** (plan 2.8 point 2): the signing credential exists nowhere else, so PRs and forks read the cache but never write it - the same boundary kache draws for its remote.
- **`!cancelled()`, not `success()`**: the toolchain closure is valid even when a test fails.
- **Only `build-and-test`** pushes: it runs on every main push; the release jobs of the same push would re-upload the identical closure.
- **Non-fatal**: a failed push shows as a failed step without reddening the build. The cache is a cold-start optimization, not a build input.

## Secret handling

`NIX_CACHE_SIGNING_KEY` reaches the script only via `env:`. Nix takes a signing key only as a file, so the step writes it to a `mktemp` file in `$RUNNER_TEMP` (pod-local, not the shared PVC), stripping any trailing newline, and a `trap` removes it on exit. S3 auth reuses `KACHE_S3_ACCESS_KEY` / `KACHE_S3_SECRET_KEY` as `AWS_*`. A warm run is a no-op: `nix copy` skips paths the bucket already holds.

## Validation

- Both edited YAML files parse; `actionlint` reports only pre-existing findings (unknown `pond-ci` label, an `SC2016` in `publish-release`).

## What only CI can prove

- The runner image's nix has `s3://` store support - check with `nix store info --store 's3://pond-nix-cache?endpoint=nbg1.your-objectstorage.com&region=nbg1'` on a runner.
- The kache credential really has write access to `pond-nix-cache`.
- The first push is a full closure upload; later ones should take seconds.

## Release note

CI now publishes the pinned toolchain closure to pond's public binary cache on every push to main, so a fresh checkout, a fork PR or a cold CI pod substitutes the toolchain instead of building it.


tenequm:
Both items under "What only CI can prove" are already proven outside CI:

- **s3-aws-auth in the image nix**: the pond-runner:10 smoke test ran the exact `nix store info --store 's3://...'` probe *inside the container* and got the AWS credential-chain error (not an unknown-scheme error) - s3 + s3-aws-auth are compiled into the static nix.
- **Write access of the kache credential to pond-nix-cache**: the cache was pre-seeded (108 paths, 5.3 GiB) using that same Hetzner credential - writes work.

The step's first main-branch run should be a fast no-op copy against the seeded cache.

Issue #266, opened 2026-09-16
Memory ceiling + supervised self-restart for mcp and serve


tenequm:
#249 shipped the allocator trim for `mcp` and `serve`, described in its own text as "a mitigation, not a cure": it returns freed pages to the OS after large transients, but it does nothing about a process whose live heap genuinely keeps growing.

This issue tracks the remaining operational guardrail from #245's worklist - the half that #249 did not cover:

- An RSS ceiling for the long-lived `mcp` and `serve` processes.
- Supervised self-restart when that ceiling is crossed, so a leak or an unknown regression degrades into a restart instead of an OOM kill.
- The earlier `oom_score_adj` idea from the same worklist: set or document it so the kernel prefers the large pond process when it does have to choose.

Refs #245, #111, #249.

tenequm:
Real-world data point (2026-09-17, observed while closing #245): a `pond mcp` process on the operator box reached 16.2 GiB RSS with a 20.1 GiB VmHWM after ~23 h serving one client session, on the pre-campaign binary (fixes merged in #245's lane are not yet deployed there). Box total is 62 GiB; the kernel's low-memory reaper started killing unrelated background processes. Exactly the failure mode this issue's ceiling + supervised restart addresses, and a deploy of the fixed binary is the first mitigation.

tenequm:
Closing as won't-do, after building it and deciding against merging (see #273).

The short version: the 16 GiB `pond mcp` that motivated this issue was a pre-campaign binary, and the #245 campaign was the actual fix - post-campaign peak RSS is in the hundreds of megabytes, and no build since has shown runaway growth. An in-process ceiling would be a backstop that never fires, paid for with ~950 lines of shutdown-path code and two config knobs.

Where each piece of this issue's ask landed:

- Memory ceiling: use the supervisor instead - `MemoryMax=` + `Restart=on-failure` on the systemd unit for `pond serve`. For `pond mcp` as a stdio child there is no supervisor, but there is also no observed need on current binaries.
- Supervised self-restart: same answer - restart policy belongs to the supervisor.
- `oom_score_adj` guidance: fine to document in the ops docs if wanted; not worth in-process code on its own.

The real memory work continues in #229 (sync cost proportional to the delta, not the store) and #257 (residual allocation backlog). Reopen only if a post-campaign binary demonstrates unbounded growth.

Refs #273 #229 #257.

Issue #111, opened 2026-07-20
pond serve leaks memory until OOM-killed (0.13.2): 157 MB -> 3.6 GB RSS over ~5 days


tenequm:
`pond serve` grows unbounded until the kernel OOM-kills it. On a 15 GB host it reached 5.5 GB RSS and was killed by the OOM killer; after an automatic restart it climbed back to 3.6 GB RSS + 3.2 GB swap over the next 4.9 days. `pond mcp` shows the same pattern independently.

## Version

```
pond 0.13.2 (cc94b7d x86_64-linux)
```

Confirmed the leaking process was running this build (binary mtime predates process start), so this is not a stale-binary artifact.

## Environment

- Ubuntu 24.04.4 LTS, kernel 6.8.0-124-generic
- Intel N100, 4C/4T, 15 GB RAM, 8 GB swap, `vm.swappiness=10`
- `pond serve --host 127.0.0.1 --port 9797` run as a systemd `--user` service with `Restart=always`, `MemoryMax=infinity`
- Mixed workload: MCP clients over a token-gated HTTP proxy, plus periodic search/ingest

## Observed

Steady-state footprint after ~5 days uptime, before any intervention:

| process | uptime | RSS | swap |
|---|---|---|---|
| `pond serve` | 4.9d | 3617 MB | 3249 MB |
| `pond mcp` | 5.9d | 2632 MB | 2258 MB |

Combined 6.13 GB RSS + 5.51 GB swap on a 15 GB box. Host-wide this left 187 MB free with swap 100 % consumed (8130/8191 MB).

The kernel had already killed `pond serve` once:

```
Wed Jul 15 21:42:55 2026  Out of memory: Killed process 1424452 (pond)
  total-vm:11879864kB anon-rss:5546588kB file-rss:2628kB shmem-rss:0kB
  UID:1000 pgtables:21200kB oom_score_adj:200
  task_memcg=/user.slice/user-1000.slice/user@1000.service/app.slice/pond-serve.service
```

`Restart=always` brought it back one minute later, and growth resumed from there. Two OOM events total in `dmesg`, four mentions across 30 days of journal.

## Effect of a restart

Restarting both processes reclaimed the memory immediately, which points at accumulation in-process rather than genuine working set:

| | before | after |
|---|---|---|
| host RAM used | 12125 MB | 5218 MB |
| host available | 3636 MB | 10544 MB |
| swap used | 8130 MB | 2282 MB |
[...]

Issue #257, opened 2026-09-16
perf(memory): residual optimization backlog after the #245 fixes


tenequm:
Residual memory-optimization backlog from the #245 profiling work (2026-09-16, instrumentation lane in #253). None of these are OOM-critical - the dominant terms are addressed by #249, #250, #251, #252, #254, #255, #256. These are the remaining measured opportunities, in priority order. Full attribution reports are archived with the instrumentation work; scenario numbers below are from the large-profile heaptrack/dhat runs.

- [x] **B6 tier 3 - chunked rowmap scan/encode window (largest remaining lever).**
  #255 landed tiers 1-2 (pre-reservation, incremental distinct, streamed serialization). Tier 3 chunks the scan-collect-encode path so scan input batches drop as they are folded, bounding the working window instead of retaining all rows' strings until encode. Path from the original 497 MiB cold-build transient toward a ~110-150 MiB peak (70-78% cut); design documented in #255's review discussion. Also applies to the `MAX_ROWMAP_DELTAS` compaction branch.

- [ ] **B6 residual - compaction while disordered bakes a permanent scan fallback.**
  If a store is compacted while its fragments are row-id disordered (e.g. `always_compact()` after scattered `merge_update`), the rewrite leaves one fragment whose live row ids straddle the others permanently. `Store::ascending_row_id_fragments` then never finds a disjoint plan, so every cold rowmap build takes the counted fallback path (`ROWMAP_SCAN_FALLBACKS` increments forever). Follow-up: have compaction restore ascending row-id order, or add a repair/reorder path. Surfaced during the #260 fix work; see the chunked-scan PR discussion.

- [ ] **B6 residual (second mechanism) - `merge_update` stops returning rewritten rows in row-id order above ~150k rows.**
  Found while landing #258: lance `merge_update` returns rewritten rows in row-id order at 100k/120k rows, but scrambled at 160k/260k/1M, with the first descent at scan index 143999 (the first rewritten fragment). Consequence: #260's ordered fast path correctly refuses at production scale after any embed pass, falling back to the chunked cold build. Distinct from the compaction-while-disordered mechanism above. Numbers: PR #258 body and the partial-embed scenario it added. Refs #258.

- [ ] **B7 leftover - dense all-null embedding buffer (155.6 MiB at 200k messages, embedder-less ingest only).**
[...]

PR #273, opened 2026-09-17
feat(serve): memory ceiling with supervised self-restart

tenequm:
## Problem

#249 shipped the allocator trim and said so itself: a mitigation, not a cure. It hands back freed pages after a transient, but a process whose live heap genuinely keeps growing still walks to an OOM kill. #111 measured `pond serve` at 3.6 GB RSS + 3.2 GB swap after 4.9 days (steady state 157 MB), #245 recorded 21 kills in 9 days across `serve`, `mcp` and `sync`, and a field `pond mcp` reached 16.2 GiB RSS / 20.1 GiB HWM after 23 h on a 62 GiB box. Worse, #245 caught the victim selection inverted: the kernel killed unrelated ~30 MB processes while the multi-GB `pond mcp` survived, because those carried the higher `oom_score_adj`.

#111 also recorded the thing that makes a guardrail possible: a restart recovers everything. So the remaining guardrail is to turn unbounded growth into a bounded restart.

## What this adds

A resident-set ceiling for the two long-lived surfaces, plus the `oom_score_adj` half of #245's worklist.

```toml
[runtime]
memory_ceiling = "4 GiB"   # default; "0" disables; floor 256 MiB
oom_score_adj  = 200       # default only when nothing else set one; 0 opts out
```

Both mirror to env (`POND_RUNTIME_MEMORY_CEILING`, `POND_RUNTIME_OOM_SCORE_ADJ`), so a container sizes the ceiling to its cgroup with no config file.

## Mechanism

Every 30 s a background task in `serve` / `mcp` reads `VmRSS` from `/proc/self/status` - the probe #245's instrumentation already landed in `memprobe.rs`, moved to `memory.rs` so a default build (not just the `mem-probe` feature) has it, and re-exported so the bench lane is unchanged. No new dependency.

Three properties make the decision trustworthy:

1. **The trim runs before the breach is believed.** An over-ceiling reading triggers `malloc_trim` and is then re-read; only the second reading counts. glibc's retained arenas are exactly the thing that can hold a process above the ceiling while it owns nothing, and restarting for reclaimable pages would be a bug.
2. **Two consecutive post-trim readings** (one check interval, 30 s, apart) before acting, so a legitimate transient - #245's 3.8 GB sync spike lasted 8 seconds - does not restart a healthy process.
3. **Drain first, then leave.** A breach closes the intake, then waits for every in-flight request *and* any running in-serve sync cycle: the sync loop holds the same counter the request handlers do, so the stop waits out a write rather than racing it, and the guard - not the serving future - is what ends the process, so the drain always runs to its own conclusion and the code is always 75.

   Closing the intake is per surface, because the two halves of "stop taking work" and "stop delivering it" must not be the same step. On HTTP the listener closes but the MCP sessions stay up until the drain reports done - cancelling the session token cuts the response stream of every tool call still running on it (rmcp streams each request-wise response under a child of that token), which is exactly the work being drained. On stdio there is no listener to close, so the tool surface itself refuses further calls with a retryable error naming the ceiling. Once the drain ends, the process waits for the transport to close - bounded, and a fixed short grace on stdio, where nothing closes on its own - so the exit never truncates the response it just finished.

   The wait is bounded at 30 s, because one request wedged on an unreachable object store would otherwise disarm the ceiling. Past that the outstanding count is logged and the process leaves anyway. That costs the unfinished tail of a cycle, never a half-written row: pond writes are append-only single commits (`lance-append-only`), so the next sync redoes exactly what did not commit.

Then the process exits `75` (sysexits `EX_TEMPFAIL` - "retry", which is precisely the right response, and distinct from every other code pond returns).

`oom_score_adj` is set to +200 (the value two field deployments confirmed) only when the inherited value is 0. systemd's `OOMScoreAdjust=`, a container runtime's setting, or an explicit `[runtime].oom_score_adj` all win; pond only fills a vacuum. A configured `0` is the opt-out and means pond never writes the file at all - forcing a 0 over a score the supervisor chose would be overruling it, and below an inherited floor the kernel refuses the write anyway. Raising the score never needs privilege, and a failure is logged rather than fatal.

Off Linux there is no cheap resident-set reading: the guard is not armed at all and says so at startup (one `info` line, so `-v` shows it). The field evidence is entirely Linux, and adding a mach `task_info` probe for a platform with no OOM killer was not worth the unsafe block.

## What the client sees

- **`serve` (HTTP).** The listener closes, in-flight responses still complete - including a `/mcp` tool call, whose session outlives the drain - and the process exits 75. A `Restart=always` unit brings it back in about a second and the next request lands on the fresh process. **Without a supervisor the server stays down** - the guide now says so, and adds the unit (`RestartForceExitStatus=75` is redundant under `Restart=always`, included so the intent is legible in a unit that restarts selectively).
- **`serve --transport stdio` and `mcp`.** The client owns the process, so the transport cannot blink invisibly. pond finishes the tool call it is answering, refuses any further call with a retryable error naming the ceiling, prints one line to stderr naming the ceiling and the reading, and exits 75; the stdio pipe closes. The agent's pond tools disconnect and reconnect on the client's next attempt (`/mcp` in Claude Code - #245 confirmed that path live when killing a large `pond mcp` by hand). The stderr line is a plain `eprintln!`, not a tracing event, so it survives whatever log filter the client set and lands in the client's MCP log.

## Design choices worth naming

- **A drained exit, not an in-process restart.** The issue allows either. An in-process reset would rebuild the store and state but reclaim nothing a leak holds outside them, and #111's evidence is specifically that the *process* boundary is what returns a grown pond to its steady state. The alternative stays available if a future leak is ever localized to the store handle - but it would be a worse default today, because it would look like a fix while leaving the arena untouched.
- **4 GiB default.** Generous next to the measured steady state (157-600 MiB) and above the largest legitimate transient on record (3.8 GiB), so only sustained growth reaches it. Deliberately not scaled to host RAM: that needs a new dependency and turns the ceiling into a number nobody can predict from config.
- **A 256 MiB floor, enforced at config load.** A ceiling under the startup peak would ship a restart loop; the load fails instead, naming the reason.
- **The check is its own task, not folded into the 30 s rowmap-refresh loop** (which already carries the periodic trim). That loop can block for minutes on a cold rowmap rebuild - which is exactly when memory peaks - so the guard would go blind at the worst moment.

## Tests

All decision logic is unit-tested against a scripted fake meter; no test reads a real resident set (the single exception asserts only that the probe exists where the platform supports it).

`packages/pond/src/memory.rs`, 17 tests: inclusive threshold, one-sample-is-not-a-breach, exactly-N-consecutive breaches, a trim that reclaims the breach clears it, a clear sample resets the count, an unmeasurable process never breaches, idle/finished/wedged drain, a drain that refuses new work without leaking a count, the stage order (sessions close only after the drain, and the exit waits for the transport), and the `oom_score_adj` precedence cases including the `0` opt-out.

Mutation-proofed, verified by actually flipping each comparison:

- `resident_bytes >= limit_bytes` -> `>`: `the_ceiling_is_the_first_value_that_is_too_much` FAILED.
- `consecutive >= self.confirmations` -> `>`: `consecutive_samples_over_the_ceiling_breach` FAILED.

`packages/pond/src/config.rs`: default / off-switch / sized ceiling, the floor refusal, the `oom_score_adj` range refusal, and the env mirror for both knobs.

`packages/pond/src/main.rs`: `a_ceiling_stop_hands_the_exit_to_the_guard` - the serve arm parks instead of returning once the ceiling has closed the intake, and reports the closed transport on its way in. Without that park the serving future returns first, `run` ends with code 0, and dropping the runtime aborts the in-serve sync cycle the drain exists to protect. Both serving arms (`serve` and its `mcp` alias) arm and park through one helper, so they cannot drift.

`packages/pond/tests/integration/transport_http.rs`: `a_served_request_leaves_nothing_in_flight` - search, get-session and ingest each release the drain when they answer, and a served server drains to `Idle`. A handler that never released its guard would hold the drain open for its whole window and stop mid-request anyway, and nothing else would catch it: the counter is invisible in every response. `a_ceiling_stop_holds_mcp_sessions_open_until_the_drain_ends` drives a real socket and a live `/mcp` stream: the server keeps serving after the listener closes and stops once the drain reports done. Mutation-proofed - collapsing the two stages back into one makes it FAIL.

`packages/pond/src/transport.rs`: `a_draining_process_refuses_new_tool_calls` - an admitted call holds the drain and releases it on the answer, and once draining the tool surface refuses with a message naming the ceiling, without leaking a count.

## Gates

Under rustc 1.98.1: `cargo fmt --check`, `cargo clippy --locked --workspace --all-targets -- -D warnings` (and the same with `--features mem-probe`), `cargo test --locked` - all green. `docs/benchmarks/*.jsonl` untouched.

Refs #266, #111, #245.

## Release note

`pond serve` and `pond mcp` now stop themselves at a memory ceiling (`[runtime].memory_ceiling`, 4 GiB default) instead of growing until the kernel OOM-kills them: in-flight work drains first, then the process exits 75 for a supervisor or MCP client to restart. Run `serve` under one.


tenequm:
Closing unmerged, by decision rather than defect - the implementation is complete and CI is green on `6ff789c`.

Rationale: the incident that motivated this (a `pond mcp` at 16.2 GiB RSS) was running a pre-campaign binary. The #245 campaign itself was the fix for that failure, and no post-campaign binary has been observed anywhere near a ceiling. An in-process guard here is a backstop for a problem the current code has not demonstrated, and every line of it is surface we would have to carry and later reconsider.

What replaces it:

- `pond serve` under systemd gets the supervisor-native equivalent: `MemoryMax=` plus `Restart=on-failure` on the unit - same guarantee, zero pond code.
- The root-cause work stays where it belongs: sync memory proportional to the delta (#229) and the residual allocation backlog (#257).

Revisit only if a post-campaign binary shows unbounded growth in the wild. The branch `feat/mcp-memory-ceiling` keeps the finished work (including the staged-shutdown fixes found in review) recoverable at `6ff789c`.

Refs #266 #229 #257.

Issue #283, opened 2026-09-17
feat(mcp): register MCP against a long-lived pond serve (M1)


tenequm:
Milestone **M1** of the read-latency campaign. Addresses the serve-topology item of #165.

## Motivation

Today's one-shot stdio MCP registration pays a cold start on every invocation:

- **267 s** cold start for a one-shot stdio MCP process (prewarm plus rowmap build, both behind the flock).
- **47-300 s** for the first FTS query in each process, cold (#165 item 2).

Both costs vanish in a resident `pond serve` (HTTP, optionally `--with-sync`): the prewarm and rowmap build happen once, and subsequent queries hit an already-warm process.

A warm resident process also keeps the local rowmap fresh, which shrinks the version-straddle windows that the warm-get fast path (M3a) fixes - the two milestones compound.

## Scope

- `pond init` MCP registration targets the HTTP `pond serve` topology instead of one-shot stdio.
- Migration story for existing stdio registrations.
- Docs.

## Acceptance

First MCP query after client start completes in **under 10 s** against the warm store.

Issue #165, opened 2026-08-21
perf(read): remote store reads - parts residency, FTS prewarm, long-lived serve topology


tenequm:
## Why

Measured over 30 days against a remote S3 store ([2608-12 read-path doc](https://github.com/tenequm/pond/blob/main/docs/researches/2608-12-read-path-where-time-goes.md)): 77% of all pond MCP calls took over 5 s; `pond_search` p50 10 s; before #141 not one `pond_get_session` or `pond_get_message` finished under 5 s. "Your own S3 bucket" is the differentiator, so the remote read path has to feel local.

#141 (v0.14.9) landed the two biggest items - rowmap-served message-id resolution and get-family pages (warm `pond_get_message` 10,900 -> 4,003 S3 GETs). This issue tracks what is left, in order of seconds recovered per unit of work.

## Remaining work

1. **Parts summaries resident** (the "5c" item): the last per-hit S3 scan in search hydration and the parts leg of gets. Serve `parts_summary` from a compact parts-meta sidecar next to the rowmap, or extend the rowmap. Target: warm `pond_search` hydration ~158 GETs -> tens.
2. **Prewarm FTS postings from the disk index cache** (bounded): the first fts query in a fresh process measures 47-300 s cold. With fts becoming the default arm (#164) this is the first thing a new session feels.
3. **One long-lived server instead of N cold processes.** `pond serve` already mounts MCP on `/mcp` and `--with-sync` folds the sync in; document registering Claude Code against the HTTP endpoint so all sessions share one warm cache, one S3 client, one index cache. Config-level, no new code; needs a docs page and an `init` option.
4. **Give the get/sql CLI paths the disk index cache** (today scoped to serve/mcp/search).
5. **Bound concurrent expensive reads per process** (small semaphore around scan-heavy handlers) so a subagent swarm degrades to queueing instead of 300 s client timeouts.
6. After #145: re-baseline and take the date-filter win (zonemap under DataFusion 54; date-filtered vector search measured ~2 min on the remote store today). One third of all searches in the 63-day trace carry a date filter and they succeed at 28-31% vs 47-48% without - part of that gap is this.

Not in scope: the sync oracle redesign (`session_last_message_ids` full aggregation, ~47 s of a ~53 s import phase) - a write-path item, tracked separately.

## Acceptance

- `serve_mem_bench --io-trace`: GETs per warm `pond_search` and `pond_get_message` recorded before/after each item.
[...]

PR #292, opened 2026-09-17
feat(mcp): register MCP against a long-lived pond serve

tenequm:
Addresses #283 (M1 of the read-latency campaign; the serve-topology item of #165).

## Problem

MCP registers today as one-shot stdio (`pond mcp`), so every client process pays the whole cold start: ~267 s for prewarm plus the rowmap build behind the flock, and 47-300 s for the first FTS query in that process. None of that work is reusable, because the process dies with the client.

A resident `pond serve --transport http --with-sync` makes all of it durable - one warm store, one embedder, one S3 client, one rowmap, shared by every client on the host and by the periodic sync.

## Design

**What registers what.** Two registrations, in this order:

1. `pond service start` registers `pond serve --transport http --with-sync --host 127.0.0.1 --port 9797 --sync-every 5` with the OS service manager. No new supervisor: it reuses `schedule.rs`'s machinery - launchd (`KeepAlive` + `RunAtLoad`), a systemd user service (`Restart=always`, `WantedBy=default.target`), or a logon-triggered Task Scheduler task running `pondw.exe` (`RestartOnFailure`, `ExecutionTimeLimit` PT0S - the sync task's PT1H would kill a server every hour). The same pins the sync registration uses are applied for the same reasons: the resolved `XDG_STATE_HOME` and config file go into the unit, so the resident sync takes the same per-host lock (with `--no-wait`, per the spec) and reads the same config as a manual run.
2. The client is pointed at that endpoint: `claude mcp add -s user --transport http pond http://127.0.0.1:9797/mcp`. Codex gets `codex mcp add pond --url ...` printed, as before - pond never edits another tool's config for it.

`pond service stop|status|logs` complete the verb; `status` recovers the backend and the live `/mcp` URL out of the registration itself, and `logs` goes to journalctl on systemd, the `serve.log` file elsewhere.

**Host allowlist.** The registration is loopback-only and refuses a non-loopback `--host`. rmcp's default `Host` allowlist is `localhost` / `127.0.0.1` / `::1` with no port pinned, and its matcher treats a portless entry as port-agnostic, so `Host: 127.0.0.1:9797` passes with nothing widened. Exposing the endpoint under another name stays an explicit `pond serve --allowed-host` deployment.

**`pond init`.** The MCP section now asks how clients should reach pond - resident HTTP (preselected) or one-shot stdio - and `--mcp-transport http|stdio` answers it non-interactively. Choosing HTTP registers the service first, then the client, so a client is never pointed at an endpoint nothing will answer; a service registration that fails (no systemd user instance, for instance) downgrades that run to stdio and names `pond service start` as the retry.

**Migration.** `init` reads the current registration out of `claude mcp get pond` (`Type: stdio` vs `Type: http` + `URL:`) and plans Keep / Add / Switch. A switch removes the old user-scope entry and writes the new one, then says so: `mcp: switched Claude Code from one-shot stdio (pond mcp) to HTTP at http://127.0.0.1:9797/mcp (user scope)`. Interactively it confirms first. Non-interactively only an explicit `--mcp-transport` switches a working registration - a bare `--yes` re-init keeps whatever topology the host has, and a fresh non-interactive host still defaults to stdio, because a `--yes` run may be sandboxed (the e2e suite drives one) and an OS service registration would outlive the sandbox. An already-correct HTTP registration still re-registers the unit interactively, which is how a re-run heals a stale unit after an upgrade - the same repair posture the schedule section has.

Docs: README quickstart plus a `pond service` block under "Run a server" (including the `loginctl enable-linger` caveat for Linux user units), and additive spec 7.8 wording on the `pond init` MCP section and a new `pond service` verb bullet.

## Test evidence

`cargo test -p pond-db --bin pond`: **66 passed, 0 failed**. `cargo clippy -p pond-db --lib --bins --tests -- -D warnings` and `cargo fmt` clean.

New unit tests:

- `init.rs`: `claude mcp get` output parses into both topologies (including an SSE entry planning a switch, and a no-`Type` body reading as absent); `mcp_plan` switches only when the shape differs (stdio -> http, and a stale port); the add argv targets the loopback endpoint and the stdio form stays byte-identical to the documented one; `pick_mcp_transport` prefers the flag, then the host's existing shape, then stdio.
- `schedule.rs`: the default endpoint is loopback and its URL is `http://127.0.0.1:9797/mcp`; endpoints round-trip out of all three registration shapes (plist `<string>` wrappers, systemd `ExecStart=`, Task Scheduler `<Arguments>`) and a registration missing a flag reports no endpoint rather than a guess; `service_start` refuses a non-loopback bind; the unix templates supervise a `--with-sync` serve with the config and state pins and floor `--sync-every` at 1; the Windows task is unbounded, logon-triggered, `RestartOnFailure`, with both pins as arguments (runs on the msvc CI leg).

Four help snapshots updated plus one new (`help_service`).

## Manual verification (not automated)

End-to-end latency needs a real warm store and a real client restart, so it stays manual:

1. `pond service start` on a host with a populated store, then `pond service status` (expect `active (systemd, http://127.0.0.1:9797/mcp)`) and `pond service logs` until the first sync cycle logs.
2. `pond init --mcp-transport http` on a host with an existing stdio registration; expect the switch line, then `claude mcp get pond` showing `Type: http` with the loopback URL.
3. Restart Claude Code and time the first `pond_search` call: the acceptance bar is under 10 s against the warm store (stdio measured 47-300 s for the first FTS query per process).
4. `pond init --mcp-transport stdio` to confirm the reverse switch, and `pond service stop` to confirm the unit and the process are gone.

## Out of scope

- `pond status` does not grow a service line; `pond service status` is the surface. Adding it means threading a second snapshot through the status JSON documents, which is a separate change.
- No readiness gate beyond a 15 s TCP probe after registration: a cold first start builds the rowmap before it binds, so a timeout is reported as "not answering yet" rather than failing the registration.
- Linux user units stop at logout unless the host has lingering enabled; `loginctl enable-linger` is documented, never invoked.
- Linux without a systemd user instance gets a named error, not a crontab fallback - cron cannot supervise a resident process.
- No automated e2e latency test (see above).

## Release note

`pond init` can now point MCP clients at one resident `pond serve` over HTTP instead of a cold `pond mcp` per client, and the new `pond service` verb keeps it alive via your OS service manager. Re-run `pond init --mcp-transport http` to switch an existing stdio registration.


tenequm:
Note for this PR's era: the straddle-delta machinery from #298 (parked branch perf/285-warm-get-fast-path, tip 18ab06f, fully validated - mid-run straddle 196 -> 11 GETs / ~6 s -> ~1.4 s) becomes structurally valuable once this long-lived serve topology lands, because a resident map with syncs committing underneath is exactly the mid-run straddle case. Cherry-pick it from the parked branch when this merges; disposition details in the #298 closing comment.

